PR #1, opened 2016-05-04
Remove CI-added compiler warning flags on regular builds

jwnimmer-tri:
This is toward RobotLocomotion/drake#1936, where these will reappear under `drake-distro/drake/CMakeLists.txt` instead.

PTAL and see if this looks like the correct starting point.

/cc @david-german-tri


jamiesnape:
LGTM.

PR #7, opened 2016-06-30
Disable libbot for sanitizer builds

mwoehlke-kitware:
Disable libbot in clang sanitizer builds to work around libbot not picking up sanitizer flags, which makes it incompatible with LCM that newly *does* pick up the sanitizer flags. Clean up the driver script. Fix permissions on `run_local.bash`.


mwoehlke-kitware:
I'm pushing this since the builds are currently broken and I'm about to take off for tonight, but please be careful; I haven't actually gotten it to build successfully yet on my machine due to various unrelated errors.

PR #9, opened 2016-07-04
CI improvements, take 2

mwoehlke-kitware:
Should look a lot like #7, but should actually *work* this time. Evidence: https://drake-cdash.csail.mit.edu/buildSummary.php?buildid=46319. The reason this was being problematic is because libbot is Just Plain Broken on Windows, and was not supposed to be enabled, but the changes to disable it for sanitizer builds accidentally were enabling it otherwise, even on Windows. For some reason, this didn't seem to always work; depending on some of the other cleanup, libbot might still be (correctly) not be built.

This does a few things:

- Simplify conditional logic in the script. Most environment variables are copied into CMake variables (simplifies later usage), and boolean options are treated as CMake booleans.
- Simplify various conditionals, either using exact string compares where appropriate or collapsing multiple conditionals into a singular regular expression conditional.
- Changes the build and install directories to remove lingering PODS-isms.
- Disables libbot where problematic (sanitizer builds, Windows builds).
- Fix YA bug in `run_local.bat`.


ctest_driver_script.cmake
```diff
@@ -5,26 +5,36 @@
...
 #   buildname             optional    value for CTEST_BUILD_NAME
 #   site                  optional    value for CTEST_SITE
 
+set(COVERAGE        $ENV{coverage})
+set(DEBUG           $ENV{debug})
+set(DOCUMENTATION   $ENV{documentation})
+set(MATLAB          $ENV{matlab})
+set(ROS             $ENV{ros})
+set(MEMCHECK        $ENV{memcheck})
+set(OPENSOURCE      $ENV{openSource})
```
jamiesnape: `OPEN_SOURCE`

ctest_driver_script.cmake
```diff
@@ -5,26 +5,36 @@
...
+set(TRACK           $ENV{track})
+
 cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
```
jamiesnape: Maybe put this above all the `set` commands.

ctest_driver_script.cmake
```diff
@@ -569,13 +581,13 @@ set(ENV{CMAKE_CONFIG_TYPE} "${DASHBOARD_CONFIGURATION_TYPE}")
 set(CTEST_CONFIGURATION_TYPE "${DASHBOARD_CONFIGURATION_TYPE}")
 set(CTEST_TEST_TIMEOUT ${DASHBOARD_TEST_TIMEOUT})
 
-if("${CTEST_CMAKE_GENERATOR}" MATCHES "Unix Makefiles")
+if("${CTEST_CMAKE_GENERATOR}" STREQUAL "Unix Makefiles")
```
jamiesnape: Maybe `CTEST_CMAKE_GENERATOR STREQUAL "Unix Makefiles"` to match the style of the other `if` commands.

mwoehlke-kitware: Indeed it should be... good catch, thanks!

jamiesnape:
Do you have a CDash submission for the latest version of this PR?

jamiesnape:
@mwoehlke-kitware All the Windows builds became unstable when I merged this.

mwoehlke-kitware:
Hmm, the test failures? Strange, looks like maybe a PATH issue?

jamiesnape:
Can you split this into three PRs?
1. All the tidying.
2. Disable libbot on Windows.
3. Everything else.

mwoehlke-kitware:
Would you object to resetting master to 5c8a13a848ff68?

jamiesnape:
Not especially, but if I do, it will trigger every continuous job on Jenkins.

mwoehlke-kitware:
I think I'd generally prefer to reset master than merge reverts when we're doing this "try, break, immediately roll back" cycle. For one it makes it cleaner to try again (e.g. I can't rebase the branch now because the history "is already merged").

The point about CI is taken, though... how about this? Next PR we try, reset locally, merge the branch locally, push that... Then the CI only triggers for the new stuff, which it would have anyway.

jamiesnape:
Sounds fine.

mwoehlke-kitware:
>> Next pr [...]

...would be #11.

PR #11, opened 2016-07-22
Disable libbot in clang sanitizer builds

mwoehlke-kitware:
Same story as #7, take 3...

Due to the (somewhat) recent LCM updates, LCM is now picking up compile flags (particularly in this case, sanitizer options), which is causing the sanitizer builds to fall apart because libbot does NOT pick up build flags, and the non-sanitized libbot build is incompatible with the sanitized LCM libraries.

Since a proper fix requires a non-trivial refactoring of the mess that is the libbot build system (and, since we want to get rid of the monolithic libbot anyway), for now at least just turn off libbot in sanitizer builds. Also, take care that libbot remains disabled on Windows (as it is currently), as the build there is broken.


jamiesnape:
Merged.

mwoehlke-kitware:
@jamiesnape, thanks! (And I see you fixed setting `PATH` in `run_local.bat`... one fewer commit from #7 to split :tada:)

PR #14, opened 2016-07-26
Use new option to skip drake build

mwoehlke-kitware:
Update CI driver to use the new mechanism for skipping the build of drake itself added in RobotLocomotion/drake#2896. This is required because the refactor removes the mechanism by which we previously caused the build to be skipped.


jamiesnape:
So a TODO is to remove `drake_ADDITIONAL_BUILD_ARGS:STRING` once https://github.com/RobotLocomotion/drake/pull/2896 is merged.

mwoehlke-kitware:
Test build:
https://drake-cdash.csail.mit.edu/buildSummary.php?buildid=60403
https://drake-cdash.csail.mit.edu/buildSummary.php?buildid=60406

(Don't mind the build/test errors, the important part is that drake itself wasn't built as part of the superbuild... if anything the build errors in drake proper make this more obvious because they aren't in the superbuild.)

mwoehlke-kitware:
> So a TODO is to remove `drake_ADDITIONAL_BUILD_ARGS:STRING` once https://github.com/RobotLocomotion/drake/pull/2896 is merged.

I *think* that's safe; CI never builds a non-merged branch? (Otherwise we might want to wait a while; the `drake_ADDITIONAL_BUILD_ARGS` stuff will be harmless with RobotLocomotion/drake#2896 merged, so there is no rush to remove it.)

jamiesnape:
> CI never builds a non-merged branch?

True in practice.

PR #15, opened 2016-07-26
Use 'build' instead of 'pod-build' for binary dir

mwoehlke-kitware:
This changes the build and install directories used by CI to ~~match the current build instructions~~ not use the old PODS directories, in order to flush out if anything was making assumptions that those were being used (assumptions which would be wrong for anyone following the current build instructions).

@jamiesnape, @david-german-tri, we might want to have a merge plan for this... it won't surprise me if it breaks something, but ultimately this patch *really* should go in, because any issues resulting from this are likely to bite ordinary users as well. Accordingly, if we can run some tests first, especially on e.g. OS X with MATLAB, we should, and we may want to plan to merge it at a time when we're standing by to try to fix any breakage ASAP rather than having to go through the merge-revert-repeat dance.

(This is also the last remaining part of #9.)

### DIFF
--- a/ctest_driver_script.cmake
+++ b/ctest_driver_script.cmake
@@ -160,7 +160,7 @@ elseif(COMPILER STREQUAL "msvc-64")
 endif()
 
 set(CTEST_SOURCE_DIRECTORY "${DASHBOARD_WORKSPACE}")
-set(CTEST_BINARY_DIRECTORY "${DASHBOARD_WORKSPACE}/pod-build")
+set(CTEST_BINARY_DIRECTORY "${DASHBOARD_WORKSPACE}/build")
 
 if(WIN32)
   if(COMPILER MATCHES "ninja")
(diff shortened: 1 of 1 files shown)


jamiesnape:
We do not have the infrastructure to test forks of CI, unfortunately.

ctest_driver_script.cmake
```diff
@@ -403,10 +403,10 @@ if(COMPILER STREQUAL "cpplint")
       "*** CTest Result: FAILURE BECAUSE CPPLINT WAS NOT FOUND")
   endif()
   set(CTEST_BUILD_COMMAND
-    "${CMAKE_CURRENT_LIST_DIR}/cpplint_wrapper.py --cpplint=${DASHBOARD_CPPLINT_COMMAND} --excludes=(\\.git|doc|pod-build|thirdParty) ${DASHBOARD_WORKSPACE}/drake")
+    "${CMAKE_CURRENT_LIST_DIR}/cpplint_wrapper.py --cpplint=${DASHBOARD_CPPLINT_COMMAND} --excludes=(\\.git|doc|build|thirdParty) ${DASHBOARD_WORKSPACE}/drake")
 endif()
 
-set(DASHBOARD_INSTALL_PREFIX "${DASHBOARD_WORKSPACE}/build")
+set(DASHBOARD_INSTALL_PREFIX "${DASHBOARD_WORKSPACE}/install")
```
jamiesnape: I thought it was normally `${CTEST_BINARY_DIRECTORY}/install`

mwoehlke-kitware:
> We do not have the infrastructure to test forks of CI, unfortunately.

Not the full suite of builds, no, but can't someone¹ log into one of the OS X instances and use the `run_local` script? Tomorrow, I can try likewise for Windows and/or Linux.

(¹ ...someone that knows *how* to log into the OS X instances, that is.)

jamiesnape:
>  log into one of the OS X instances

They are not publicly accessible.

ctest_driver_script.cmake
mwoehlke-kitware: Er... right. Actually, using non-default directories is possibly a good thing, as it will help ensure we aren't making assumptions. (The important part of the change isn't actually to match the defaults, because if it *only* works for the defaults, that's still a bug. The *important* part is to stop using the old PODS defaults in order to flush out if we're depending on those in any way. In that sense, doing something a little crazy like sticking a timestamp in the build and install dirs might not be a bad idea...)

jamiesnape: It has been requested that CI matches the defaults, so no-craziness unfortunately.

jamiesnape:
All is quiet and there are a lot of hot instances.

jamiesnape:
Reverted.

jamiesnape:
Broke Windows and broke documentation publication.

jamiesnape:
OS X and Linux other than the documentation publication seem fine.

jamiesnape:
Fix the paths in [ctest_publish_documentation.bash](https://github.com/RobotLocomotion/drake-ci/blob/master/ctest_publish_documentation.bash).

mwoehlke-kitware:
> Fix the paths in `ctest_publish_documentation.bash`.

Done. For Windows, maybe wait until I can run a test build? :smile:

mwoehlke-kitware:
It looks like there is something else in the driver script that monkeys with what is now the build directory (argh!!). The tests fail because all the executables are, well, AWOL...

PR #17, opened 2016-07-27
Use 'build' instead of 'pod-build' for binary dir (take 2)

mwoehlke-kitware:
See #15. Test windows build: https://drake-cdash.csail.mit.edu/buildSummary.php?buildid=61733 (at least some of the test failures are likely due to firewall issues, but the systemic problem is noticeably fixed).


mwoehlke-kitware:
(from #15)
> It looks like there is something else in the driver script that monkeys with what is now the build directory...

...which I fixed in this version by using `DASHBOARD_INSTALL_PREFIX` instead of hard-coding the directory in another place. (Which would have prevented the problem from happening in the first place...)

PR #18, opened 2016-07-27
Add WITH_IPOPT

jwnimmer-tri:
Add IPOPT setting and passthroughs, and duplicate the "don't run on APPLE" logic currently in the superbuild.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/18)


jwnimmer-tri:
Better documentation like RobotLocomotion/drake#2978 would have helped this get correct the first time around.

PR #22, opened 2016-08-24
Initial rough cut at refactoring CI driver

mwoehlke-kitware:
Start refactoring the monolithic CI driver script into logical parts. This should eventually make the script easier to follow, and should also make it easier to tweak for different operations (e.g. the cpplint and ROS builds).

This is an initial rough cut consisting only of mass moving large chunks of the script to separate files. More splitting and refactoring remains to be done, but hopefully getting this larger piece done first will help with review and conflict resolution while the refactoring is ongoing.


jamiesnape:
Is this literally a split or are there other changes?

mwoehlke-kitware:
Reviewing notes... I've been working on a more complicated version of this that includes quite a few files and has been incorporating various changes in the process — i.e. I have a plan for how the directory structure is working out so far ­— and then noticed that master had changed a bunch in the mean time. This is an attempt to start moving things into place in order to reduce conflicts as I'm working, and also to produce something that is hopefully easier to review.

If I did this right, the new files are "mostly" mass-moves of logic from the current monolithic script, with very little change in logic order. (`packages.cmake` has been refactored more completely, though the extent of the change is just using some helper macros to cut down duplication; the resulting logic should be unchanged.)

mwoehlke-kitware:
> Is this literally a split or are there other changes?

See previous comment (that I was writing at the same time you wrote the above :smile:).

A good way to review this would be to check it out locally and copy/paste the contents of the new files over where said files are `include`d, then compare the result against master. The diff should be minimal, besides one or two blocks setting up the build ID that got rearranged in the process. (AFAIK these shouldn't be order dependent...)

mwoehlke-kitware:
To expand on the previous answer, aside from said reordering, there should be no logic changes (if I did it right, and if nothing changes again in the interim).

PR #24, opened 2016-08-26
Initial rough cut at refactoring CI driver

mwoehlke-kitware:
Start refactoring the monolithic CI driver script into logical parts. This should eventually make the script easier to follow, and should also make it easier to tweak for different operations (e.g. the cpplint and ROS builds).

This is an initial rough cut consisting only of mass moving large chunks of the script to separate files. More splitting and refactoring remains to be done, but hopefully getting this larger piece done first will help with review and conflict resolution while the refactoring is ongoing.

PR #25, opened 2016-08-29
CI Refactor, part 2

mwoehlke-kitware:
Continue refactoring the CI script. This:

- Moves some repetitive error reporting code to a helper function.
- Moves the AWS logic to a separate file.
  - Refactors the same to use helper functions for creating temporary files and setting file permissions.
  - Uses `string(RANDOM)` instead of `ENV{RANDOM}` for creating temporary files (on Windows).
- Simplifies and makes generic the mechanism for cleaning up temporary files.
- Moves ancillary scripts to their own subdirectory.
- Moves the logic to set up the site and build identification to a separate file.
- Moves the logic to extract the compiler from the environment to (near) the beginning of the script, for consistency with the logic that extracts the rest of the configuration from the environment.
- Moves some other general setup (source/binary/install dirs, CTest git/update commands) logic much closer to the beginning of the script.

Unlike #24, the last three items above involve some actual reordering of logic, although I don't believe any data dependencies were violated by the changes made. The remaining changes — with the exception of `string(RANDOM)` vs. `ENV{RANDOM}` — should be strictly logic-preserving.


mwoehlke-kitware:
I *think* I cut this at a point where the changes are still relatively easy to follow, even looking at the cumulative diff. However, I also tried to keep the individual commits manageable.

jamiesnape:
Error messages need to begin `*** CTest Result: FAILURE BECAUSE`.

mwoehlke-kitware:
Ack, lost the `***` by accident... good catch, thanks!

jamiesnape:
Can you fix the merge conflict?

jamiesnape:
LGTM.

mwoehlke-kitware:
Bah; syntax error... reverted. Might wait until tomorrow to try again.

PR #26, opened 2016-09-14
Use drake build as set up by superbuild

mwoehlke-kitware:
Fix location of drake-proper to match the build that is set up by the superbuild. Remove logic to nuke the drake-proper build before building it and to rewrite the cache from scratch, so that we instead use the build that was set up and configured by the superbuild, as normals users would.


jamiesnape:
Are these passed down to drake?
`LONG_RUNNING_TESTS`
`TEST_TIMEOUT_MULTIPLIER`

jamiesnape:
Also will the configure be displayed correctly on the Drake CDash dashboard?

mwoehlke-kitware:
Local experimental builds:

https://drake-cdash.csail.mit.edu/buildSummary.php?buildid=115368
https://drake-cdash.csail.mit.edu/buildSummary.php?buildid=115369

mwoehlke-kitware:
Reverted

PR #27, opened 2016-09-22
Use drake build as set up by superbuild

mwoehlke-kitware:
Fix location of drake-proper to match the build that is set up by the superbuild. Remove logic to nuke the drake-proper build before building it and to rewrite the cache from scratch, so that we instead use the build that was set up and configured by the superbuild, as normals users would.

This is #26, take 2, hopefully with the glitches fixed...


mwoehlke-kitware:
Local tests: [drake-superbuild](https://drake-cdash.csail.mit.edu/buildSummary.php?buildid=122645), [drake](https://drake-cdash.csail.mit.edu/buildSummary.php?buildid=122657)

jamiesnape:
LGTM.

PR #28, opened 2016-10-14
Add Xenial builds to the CTest driver script

BetsyMcPhail:


BetsyMcPhail:
@jamiesnape  and @mwoehlke-kitware  PTAL. I am not sure if the set-up script is being called in the best way... Suggestions/comments please.


One other issue, drake requires CMake 3.5 while the ctest_driver_script requires CMake 3.6. I think the solution is to install CMake 3.6 on the xenial AMI. Then the install script can check if CMake is already installed before installing the default version (3.5). And, of course, CMake needs to be installed on the image before it can run this script!

jamiesnape:
Yes, though you will need to install CMake 3.7 on the AMI once it is out. While we are still testing clone the latest release branch, build from source, and install to `/usr/local/bin` (i.e., the default install location).

ctest_driver_script.cmake
```diff
@@ -251,6 +262,10 @@ if(COMPILER MATCHES "^scan-build")
   file(MAKE_DIRECTORY "${DASHBOARD_CCC_ANALYZER_HTML}")
 endif()
 
+if(COMPILER MATCHES "^xenial")
+  execute_process(COMMAND bash "-c" "yes | sudo ${DASHBOARD_WORKSPACE}/setup/ubuntu/16.04/install_prereqs.sh")
+endif()
```
jamiesnape: You need to check the return value and fail if the script fails.

BetsyMcPhail:
PR for set-up script changes: [#3806](https://github.com/RobotLocomotion/drake/pull/3806)

The latest release branch of CMake is installed on the Xenial AMI from today (2016-10-24).

jamiesnape:
Can you add support for an environment variable `provision` that is set to `true` or `false`?

jamiesnape:
Do the values of the `CC`, `CXX`, `FC`, `F77` environment variables need to be changes for Xenial?

ctest_driver_script.cmake
```diff
@@ -251,6 +264,21 @@ if(COMPILER MATCHES "^scan-build")
   file(MAKE_DIRECTORY "${DASHBOARD_CCC_ANALYZER_HTML}")
 endif()
 
+if (PROVISION)
```
jamiesnape: `if(`

PR #31, opened 2016-10-24
Increase timeouts 4x for memcheck builds

david-german-tri:
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/31)


jamiesnape:
So 16x in total?

david-german-tri:
Yeah.  I'm not sure that's actually enough for all the tests, but anything that overruns 16x we should either shorten or suppress.

PR #32, opened 2016-10-26
Split superbuild into pre- and post-drake

mwoehlke-kitware:
Store build name in `DASHBOARD_BUILD_NAME` instead of `CTEST_BUILD_NAME` so we can set a different build name per stage. Add a new post-drake stage that rebuilds the superbuild (with `SKIP_DRAKE_BUILD=OFF`) after building drake, in order to build anything that depends on drake (e.g. director).


mwoehlke-kitware:
So... the way this works with jenkins is the "View superbuild in CDash" link will now show *two* builds, as opposed to the former single build. The two builds will be labeled "-pre-drake" and "-post-drake". Also, the build might fail in (e.g.) "BUILD POST-DRAKE SUPERBUILD", after drake proper has been built. Otherwise, things should generally work as before.

See [here](https://drake-cdash.csail.mit.edu/index.php?project=drake-superbuild&showfilters=1&filtercount=2&showfilters=1&filtercombine=and&field1=label&compare1=61&value1=jenkins-unix-gcc-experimental-release-6084&field2=buildstarttime&compare2=84&value2=now) for an example from local testing.

patmarion:
Cool!

By the way, let me know if you want help debugging that Qt error.  I'd start with asking for the output of `qmake --version` on that system.

Do you want to try running director tests, too?

    ctest_build(BUILD "${CTEST_BINARY_DIRECTORY}/externals/director/src/director-build" INCLUDE_LABEL core ...)

(note, on director/master you don't need to specify the test label)

jamiesnape:
LGTM, but we need a quieter time to merge. We should switch on director tests once this has run a day or two first.

mwoehlke-kitware:
> By the way, let me know if you want help debugging that Qt error.

IIRC it's a known problem with PythonQt... PythonQt needs to be fixed. I'm pretty sure there is a branch floating around somewhere that contains a fix, possibly even in [my fork](https://github.com/mwoehlke-kitware/pythonqt). (I've built it successfully before, but not within director. My guess is it needs mwoehlke-kitware/pythonqt@821dbdc28cfc.)

> I'd start with asking for the output of `qmake --version` on that system.

```
$ qmake-qt4 --version
QMake version 2.01a
Using Qt version 4.8.7 in /usr/lib64
```

patmarion:
looks ready to merge? @jamiesnape what do you have in mind for quieter time?

mwoehlke-kitware:
Merged; there was only one executing job and nothing queued, seemed quiet enough...

PR #33, opened 2016-11-01
Separate compiler and generator

mwoehlke-kitware:
Remove all of the `-ninja` compiler stuff and instead add a separate environment variable to communicate what generator should be used. This reduces the number of compiler options by nearly half, since all but `cpplint` came as 'plain' and `-ninja` variants.

Note that Ninja is now default when `run_local.bash` is used. Also, some straggling MSVC stuff has been removed.

@jamiesnape, the ninja jobs will need `generator=ninja` added to their environment. Otherwise it should be okay to leave `compiler` alone for now; anything after the name will be ignored. It should be possible to modify the jobs first (since it is an addition), then merge this, then later we can clean up the `compiler` settings.


ctest_driver_script.cmake
```diff
@@ -51,6 +45,12 @@ if(NOT DEFINED ENV{compiler})
 else()
   set(COMPILER $ENV{compiler})
 endif()
+if(NOT DEFINED ENV{generator})
+  message(WARNING "*** ENV{generator} was not set")
```
jamiesnape: Make this informational and say that we are defaulting to Unix Makefiles.

jamiesnape: Also does cpplint use either?

mwoehlke-kitware: Yes, actually; the download of google_styleguide is still driven by "the usual processes". (Technically I guess this only applies to the superbuild; the drake-proper build directly runs the wrapper script.)

ctest_driver_script.cmake
```diff
@@ -40,9 +34,9 @@ set(DASHBOARD_TEMPORARY_FILES "")
 
 include(${DASHBOARD_DRIVER_DIR}/functions.cmake)
 
-# Set default compiler (if not specified) or copy from environment
+# Set default compiler and generator (if not specified) or copy from environment
 if(NOT DEFINED ENV{compiler})
-  message(WARNING "*** ENV{compiler} was not set")
+  message(WARNING "*** ENV{compiler} was not set; defaulting to 'gcc'")
```
jamiesnape: Message is now incorrect.

PR #34, opened 2016-11-04
Refactor cpplint build

mwoehlke-kitware:
Separate and refactor cpplint logic into its own configuration driver script, further broken down into step scripts. Add additional helper functions for some common operations. Remove cpplint conditional logic from the generic configuration driver and, as much as possible, from the root driver script.

PR #35, opened 2016-11-09
Platform and compiler refactor

mwoehlke-kitware:
Select compiler versions based on the actual platform, and remove platform specification in `ENV{compiler}`. Make provisioning (potentially) work on any platform. Remove some minor cruft. Separate more logic into separate files. Separate generic configuration driver into step scripts and apply other improvements started in #34.

This should make it possible to run local tests on non-Ubuntu platforms without modifying the driver to remove version suffixes from the compiler. This should also be the last major refactoring prior to subprojects.

This will require that any jobs still including `-ninja` in the `ENV{compiler}` stop doing so prior to merge (since #33 it isn't needed or used), as some checks have been changed from `MATCHES` to `STREQUAL`. It will also require that the Xenial jobs remove `xenial-` from `ENV{compiler}`, for similar reasons. (Since these are all nightly rather than continuous, it should be possible to make this change post-merge before the jobs run, such that no broken builds are submitted.)


driver/compiler.cmake
```diff
@@ -0,0 +1,46 @@
...
+  endif()
+endif()
+
+# Select appropriate compiler(s)
+set(ENV{F77} "gfortran${DASHBOARD_GNU_COMPILER_SUFFIX}")
+set(ENV{FC} "gfortran${DASHBOARD_GNU_COMPILER_SUFFIX}")
+if(COMPILER MATCHES "^gcc")
+  set(ENV{CC} "gcc${DASHBOARD_GNU_COMPILER_SUFFIX}")
+  set(ENV{CXX} "g++${DASHBOARD_GNU_COMPILER_SUFFIX}")
+elseif(COMPILER MATCHES "^(clang|(include|link)-what-you-use)")
```
jamiesnape: `^(clang|(include|link)-what-you-use)$` ?

driver/compiler.cmake
```diff
@@ -0,0 +1,46 @@
...
+  set(ENV{CC} "clang${DASHBOARD_CLANG_COMPILER_SUFFIX}")
+  set(ENV{CXX} "clang++${DASHBOARD_CLANG_COMPILER_SUFFIX}")
+elseif(COMPILER MATCHES "^scan-build")
```
jamiesnape: `STREQUAL` ?

driver/compiler.cmake
```diff
@@ -0,0 +1,46 @@
...
+  else()
+    set(DASHBOARD_GNU_COMPILER_SUFFIX "-5")
+    set(DASHBOARD_CLANG_COMPILER_SUFFIX "-3.9")
+  endif()
+endif()
+
+# Select appropriate compiler(s)
+set(ENV{F77} "gfortran${DASHBOARD_GNU_COMPILER_SUFFIX}")
+set(ENV{FC} "gfortran${DASHBOARD_GNU_COMPILER_SUFFIX}")
+if(COMPILER MATCHES "^gcc")
```
jamiesnape: `STREQUAL` ?

mwoehlke-kitware: Sure... I wasn't trying to aggressively switch everything over to strict matching, but it makes sense to do so. Will push a follow-up commit for this and similar.

PR #38, opened 2016-11-15
Platform refactor

mwoehlke-kitware:
Make provisioning (potentially) work on any platform, and move platform setup to a separate file. This makes the distribution type and version available to the driver, and is the first step at supporting distributions other than Ubuntu (not officially, but at least fixing the driver to not crash and burn immediately), and to not encoding platform information in `ENV{compiler}`.

This is the first part of a second attempt to bring in the changes from #35 in smaller pieces.


mwoehlke-kitware:
Test build: https://drake-jenkins.csail.mit.edu/job/linux-clang-ninja-experimental/28

driver/platform/unix.cmake
```diff
@@ -0,0 +1,43 @@
...
+endif()
+
+# Set (non-Apple) paths for MATLAB
+if(MATLAB AND NOT APPLE)
+  prepend_path(PATH /usr/local/MATLAB/R2015b/bin)
+endif()
+
+# Get distribution information
+if(APPLE)
+  set(DASHBOARD_UNIX_DISTRIBUTION "OS X")
```
jamiesnape: If `CMAKE_SYSTEM_VERSION` is less than `16.0.0`, otherwise "macOS". Or you could just say "Darwin".

PR #41, opened 2016-11-15
Platform refactor, take 2

mwoehlke-kitware:
Make provisioning (potentially) work on any platform, and move platform setup to a separate file. This makes the distribution type and version available to the driver, and is the first step at supporting distributions other than Ubuntu (not officially, but at least fixing the driver to not crash and burn immediately), and to not encoding platform information in `ENV{compiler}`.

This is the first part of a second attempt to bring in the changes from #35 in smaller pieces, and the second attempt at #38.


mwoehlke-kitware:
Test build: https://drake-jenkins.csail.mit.edu/job/mac-clang-ninja-experimental/155 (was failing immediately before fixing the bug in `prepend_path`)

PR #48, opened 2016-11-23
More fixes for documentation

mwoehlke-kitware:
Rename internal variables for documentation as per changes to the cache names made in 23efc875dfae, which also fixes reporting (broken by the same commit). Move setting documentation related cache entries to only Drake proper, as they aren't needed in the superbuild, and only when documentation is enabled (fixing a warning about an unused user specified cache entry `BUILD_DOCUMENTATION_ALWAYS`).


driver/configurations/generic/step-drake.cmake
```diff
@@ -57,6 +55,12 @@ foreach(DRAKE_CACHE_VAR ${DRAKE_CACHE_VARS})
   endif()
 endforeach()
 
+if(DASHBOARD_ENABLE_DOCUMENTATION)
+  list(APPEND DRAKE_CONFIGURE_ARGS
+    "-DBUILD_DOCUMENTATION=ON"
```
jamiesnape: `ENABLE_DOCUMENTATION`, also should be `OFF` otherwise.

mwoehlke-kitware: `OFF` is not the default?

PR #52, opened 2016-12-01
Add Bazel build

jamiesnape:


driver/configurations/bazel.cmake
```diff
@@ -0,0 +1,73 @@
...
+  "--output_user_root=${CTEST_BINARY_DIRECTORY}")
+set(DASHBOARD_BAZEL_BUILD_OPTIONS "--compilation_mode")
+
+if(DEBUG)
+  set(DASHBOARD_BAZEL_BUILD_OPTIONS "${DASHBOARD_BAZEL_BUILD_OPTIONS}=dbg")
+else()
+  set(DASHBOARD_BAZEL_BUILD_OPTIONS "${DASHBOARD_BAZEL_BUILD_OPTIONS}=opt")
+endif()
+
+report_configuration("
```
mwoehlke-kitware: I'm not sure this shouldn't share the common case, but I'll look closer at that later...

driver/configurations/bazel.cmake
```diff
@@ -0,0 +1,73 @@
+if(UNIX_DISTRIBUTION STREQUAL "Ubuntu" AND UNIX_DISTRIBUTION_VERSION STREQUAL "16.04")
```
mwoehlke-kitware: BTW, existing logic uses version compares...

driver/configurations/bazel.cmake
mwoehlke-kitware: Never mind, I see that what is reported here is fairly minimal, similar to the cpplint case...

PR #53, opened 2016-12-01
Nuke JAVA_HOME in bazel builds

mwoehlke-kitware:
At least my machines don't seem to have `JAVA_HOME` set, suggesting that Java (at least, the system-provided version) ought to work without it. Rather than fiddling with platform-specific values, just nuke it entirely, unconditionally. (This also avoids that the condition was using the wrong names for the distribution variables.)


jamiesnape:
We probably do not need to report the value of JAVA_HOME now either.

PR #58, opened 2016-12-09
Run tests for superbuild

mwoehlke-kitware:
Modify the generic post-drake step to also run the tests for the superbuild. This will run tests for any externals that export their tests to the superbuild (which, for the moment, does NOT include Drake itself, as we don't want to run those tests twice).


driver/configurations/generic.cmake
```diff
@@ -103,8 +103,14 @@ if(NOT DASHBOARD_FAILURE)
     endif()
   endif()
 
-  if(DASHBOARD_TEST AND NOT DASHBOARD_TEST_RETURN_VALUE EQUAL 0)
-    append_step_status("TEST" UNSTABLE)
+  if(DASHBOARD_TEST)
+    if (NOT DASHBOARD_TEST_RETURN_VALUE EQUAL 0)
```
jamiesnape: `if(`

mwoehlke-kitware:
[linux-clang-ninja](https://drake-jenkins.csail.mit.edu/job/linux-clang-ninja-experimental/44/) test build

PR #60, opened 2016-12-09
Fix variable scope problem

mwoehlke-kitware:
Fix problem in `begin_stage`, introduced when `ctest_start` moved to the same (a955bc1a2cbe), where `ctest_start` doesn't have the dashboard information due to the relevant variables only being set in the parent scope.


mwoehlke-kitware:
[linux-clang-ninja](https://drake-jenkins.csail.mit.edu/job/linux-clang-ninja-experimental/45/) test build

PR #68, opened 2016-12-21
Fix cache conversion (again)

mwoehlke-kitware:
Avoid leaking literal newline characters into the CMake options. Curiously, this works at the CMake level, but results in generating make files that are not syntactically valid.


jamiesnape:
I think I ran into that problem once. I forget where and when.

PR #71, opened 2017-01-11
Ignore CTest-reported errors from bazel build

mwoehlke-kitware:
Modify CI script to only consider the exit code from bazel when determining success or failure.


jwnimmer-tri:
How's this fixup going?  I'm trying to decide if I should wait for it later today, or improve my work-around in RobotLocomotion/drake#4737 to be satisfactory for master.

mwoehlke-kitware:
See https://github.com/RobotLocomotion/drake-ci/pull/71 (untested; I'm seeing `clang-3.9: error: unsupported option '-pass-exit-codes'` trying to test it against master, which I don't believe is caused by the PR...)

jwnimmer-tri:
Re: `-pass-exit-codes`.  For now, you shouldn't set `CXX=clang-3.9` when trying to run with Bazel.  Let it use GCC.

jwnimmer-tri:
:lgtm: insofar as the plain reading of the logic change matches what I think we want.  If there are Dashboard-specific subtleties, I can't speak to those.

jamiesnape:
We just need to check that a failing build actually fails.

jwnimmer-tri:
RobotLocomotion/drake#4717 might work for that?

mwoehlke-kitware:
RobotLocomotion/drake#4717 appears to build cleanly

jamiesnape:
That is not good.

mwoehlke-kitware:
Hey, don't blame me; my change can't have made bazel itself build it successfully.

I can confirm that `run_local.bash` registers a build failure when I artificially introduce one.

mwoehlke-kitware:
Note: by "appears to build cleanly" I don't mean just that CI passed, I mean I looked at the log and AFAICT _there are no build errors_. So, it is not surprising that CI passed.

jamiesnape:
But it did not build successfully.
```
Error(s) when building project
17:18:11    1 Compiler errors
17:18:11    0 Compiler warnings
```

mwoehlke-kitware:
Okay, that's strange. I got:

```
Error(s) when building project
   2 Compiler errors
   1 Compiler warnings

  ------------------------------------------------------------------------------
  *** CTest Result: FAILURE DURING BAZEL BUILD
  ------------------------------------------------------------------------------
```

mwoehlke-kitware:
Also... *where is the error*? I see "1 Compiler errors", but I *don't see an actual error*...

On my build, I see a big honking obvious build error:

```
ERROR: /home/ubuntu/workspace/repo/drake/drake/multibody/BUILD:132:1: C++ compilation of rule '//drake/multibody:rigid_body_collision_clique_test' failed: linux-sandbox failed: error executing command /home/ubuntu/workspace/unix-gcc-experimental/_bazel_ubuntu/b0e765c79189e4b84ee4e5e4e90c1ddd/execroot/drake/_bin/linux-sandbox ... (remaining 88 argument(s) skipped).
drake/multibody/test/rigid_body_tree/rigid_body_collision_clique_test.cc:23:1: error: expected ';' before 'using'
 using std::unique_ptr;
 ^
```

jwnimmer-tri:
Let me know if I can do anything to help, but at this point it seems like its in your hands.

Note that #4717 has been revised, so it is no longer a good fail-candidate.  I guess you'll have to open a new PR with a deliberate error?

Note for local testing -- if your git checkout also has a completed CMake build in the standard location, the "drake-visualizer not found" warning from the Bazel build will not appear, and that is the warning which CTest was promoting to an error.  In CI this doesn't matter, but locally perhaps it could be misleading when attempting to repro CI.

mwoehlke-kitware:
> Let me know if I can do anything to help

Can you verify if (linux-gcc-bazel-experimental-temp#3)[https://drake-jenkins.csail.mit.edu/job/linux-gcc-bazel-experimental-temp/3/] is an expected result? CI thinks it succeeded, I suspect bazel returned `0`, and I don't see any _actual_ error, but bazel reported "1 Compiler errors".

>  I guess you'll have to open a new PR with a deliberate error? Note for local testing [...]

By "local" I mean "not driven by Jenkins"; I am testing on an AMI which I believe is/was clean. I am confident that a genuinely failing build _may_ fail. The question is whether (linux-gcc-bazel-experimental-temp#3)[https://drake-jenkins.csail.mit.edu/job/linux-gcc-bazel-experimental-temp/3/] is a genuinely failing build that doesn't show in CI as failed (which would be a problem), or if the "error" is spurious and it is succeeding correctly, in which case we are probably okay.

jwnimmer-tri:
> Can you verify if linux-gcc-bazel-experimental-temp#3 is an expected result? CI thinks it succeeded, I suspect bazel returned 0, and I don't see any actual error, but bazel reported "1 Compiler errors".

Are you sure that Bazel is emitting that message?  I think it's actually CTest needing some safe space because it saw a number surrounded by colons.

mwoehlke-kitware:
> Are you sure that Bazel is emitting that message?

No. @jamiesnape? (If it's from CTest, that would tend to indicate that this is a correct fix...)

jamiesnape:
Not sure, to be honest. CTest reports "1 compiler error" when the return value is non-zero too.

jwnimmer-tri:
Can you just open a new Drake PR that introduces a syntax error in `common/something.h`, and then make sure the build fails?  (And of course confirm that RobotLocomotion/drake#4750 still passes.)

mwoehlke-kitware:
https://drake-jenkins.csail.mit.edu/job/linux-gcc-bazel-experimental-temp/4/

PR #76, opened 2017-01-19
Add --keep_going to the Bazel build.

david-german-tri:


david-german-tri:
I can run CI on this somehow, right?

david-german-tri:
> I can run CI on this somehow, right?

From Slack: Nope.

PR #77, opened 2017-01-24
"Fix" CTest warnings

mwoehlke-kitware:
At some point, CTest started complaining about unset variables when it exits. This is because the degree of modularization interferes with the logic that CTest uses to detect an old-style "declarative" script (which we are not). Avoid this by explicitly setting the variable that CTest uses to identify "declarative" scripts.


jamiesnape:
https://github.com/RobotLocomotion/drake-ci/blob/master/ctest_driver_script.cmake#L30

mwoehlke-kitware:
Hmm, I missed that; did it fix the warnings? (My comment is prettier :stuck_out_tongue:)

jamiesnape:
Yes as far as I can tell.

PR #112, opened 2019-02-01
Add python 3 bindings to binary packages

jamiesnape:


BetsyMcPhail:
Is this expected to be 2 CDash builds or just 1?

The 'CTEST_BUILD_NAME' is set to 'DASHBOARD_JOB_NAME' at [the start of the build step](https://github.com/RobotLocomotion/drake-ci/blob/9f9e53d1c4d2012e609cd6ce0f9745f5cfd3eb3f/driver/functions.cmake#L292) . 

The CDash label and link will use the original name, as defined by '[DASHBOARD_LABEL](https://github.com/RobotLocomotion/drake-ci/blob/b7776deb361b4a12eb237f4572c8254c65f916c0/driver/site.cmake#L55)' 

It seems like the CDash setup code should get moved out of the build step?

jamiesnape:
Should be two CDash builds sharing the same label.

PR #123, opened 2021-04-22
For docs, build //doc/... lieu of manual_binaries

jwnimmer-tri:
Our deployment binaries are no longer marked manual.  This will allow us to remove their grouping from the BUILD rules, and also allows for non-manual doc tests to run in CI as part of this configuration.


jamiesnape (approved):
CI is backed up at the moment. Once that clears I will merge.

PR #129, opened 2021-11-30
Add wheel CI build

mwoehlke-kitware:
Add new CI configuration "wheel", which runs the build-wheels script to build and test Drake wheels. Tweak the wrapper script to not launch CTest in a newly spawned SSH agent when run via `run_local.bash`, since the intent is to reuse the interactive user's agent in that case.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129)


jwnimmer-tri (changes requested):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @BetsyMcPhail and @mwoehlke-kitware)

___
*[driver/configurations/wheel.cmake, line 72 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-MpmnUx97g89OPlU6fE9:-MpmnUx97g89OPlU6fEA:b-y9ct1m) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/60e923682fe2b35789fac4509783c572ceecd28a/driver/configurations/wheel.cmake#L72)):*
> ```CMake
> if(GIT_REV_PARSE_RESULT_VARIABLE EQUAL 0)
>   set(DASHBOARD_DRAKE_VERSION
>     "0.${DATE}.${TIME}+${GIT_REV_PARSE_OUTPUT_VARIABLE}")
> ```

As I understand it, this is the logic for how our PR builds and nightly builds would automatically compute "snapshot" version numbers, based on the current date.  If I'm reading this correctly, this code will produce something like `0.2021.12.01.09.05.11+deadbeef`?

(1) I don't think the snapshot builds should identify as "newer" than any official releases.  We don't want `0.2021.etc` to trump next month's `0.37.0`.  I think the snapshots should start with `0.0.2021.etc` for that reason (i.e., the date is the _patch_ revision, not the _minor_ revision).

(2) For the existing packaging builds, the version label is `drake-${DATE}${TIME}-${GIT_REV_PARSE_OUTPUT_VARIABLE}"` per `step-create-package-archive.cmake`.  Why should the wheels be different (i.e., with all of the extra dots in the middle)?  I would expect something like `0.0.202112010905-0123456789012345678901234567890123456789` or ``0.0.202112010905+0123456789012345678901234567890123456789` here?

BetsyMcPhail (commented):
To help verify this didn't introduce any breaking changes into existing CI, I started two jobs: 
https://drake-jenkins.csail.mit.edu/view/Experimental/job/linux-bionic-clang-bazel-experimental-everything-release/11336
https://drake-jenkins.csail.mit.edu/view/Experimental/job/linux-focal-clang-bazel-experimental-everything-release/9/

:lgtm: assuming those pass.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-:-MpqiRiv0g8GcJLZgM1g:b-oybo6e)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri and @mwoehlke-kitware)

mwoehlke-kitware (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[driver/configurations/wheel.cmake, line 72 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-MpmnUx97g89OPlU6fE9:-MpqTOMF5fpdDewDwXfZ:ble8sqf) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/60e923682fe2b35789fac4509783c572ceecd28a/driver/configurations/wheel.cmake#L72)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

As I understand it, this is the logic for how our PR builds and nightly builds would automatically compute "snapshot" version numbers, based on the current date.  If I'm reading this correctly, this code will produce something like `0.2021.12.01.09.05.11+deadbeef`?

(1) I don't think the snapshot builds should identify as "newer" than any official releases.  We don't want `0.2021.etc` to trump next month's `0.37.0`.  I think the snapshots should start with `0.0.2021.etc` for that reason (i.e., the date is the _patch_ revision, not the _minor_ revision).

(2) For the existing packaging builds, the version label is `drake-${DATE}${TIME}-${GIT_REV_PARSE_OUTPUT_VARIABLE}"` per `step-create-package-archive.cmake`.  Why should the wheels be different (i.e., with all of the extra dots in the middle)?  I would expect something like `0.0.202112010905-0123456789012345678901234567890123456789` or ``0.0.202112010905+0123456789012345678901234567890123456789` here?
</blockquote></details>

1) Done.

2) PEP 440 [specifies](https://www.python.org/dev/peps/pep-0440/#final-releases) a format for date-based release numbers as "Y.M" (no mention of day). "Ideally", we would use epochs to separate, but that would make "normal" versions e.g. 1!0.36, which... doesn't seem ideal :slightly_smiling_face:. But we can at least follow the recommendation of separating the various bits rather than using 14-digit numbers. (Also, parsers might choke on those, since they don't fit in 32-bits, and version numbers don't usually contain components larger than 32-bit numbers.) Also, I believe hyphens are not allowed at all; plus is the correct separator to allow a "Local version identifiers".

jwnimmer-tri (approved):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @jwnimmer-tri)

___
*[driver/configurations/wheel.cmake, line 72 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-MpmnUx97g89OPlU6fE9:-Mpr23mI0I9uTTRUJBk8:b-u04s7z) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/60e923682fe2b35789fac4509783c572ceecd28a/driver/configurations/wheel.cmake#L72)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

1) Done.

2) PEP 440 [specifies](https://www.python.org/dev/peps/pep-0440/#final-releases) a format for date-based release numbers as "Y.M" (no mention of day). "Ideally", we would use epochs to separate, but that would make "normal" versions e.g. 1!0.36, which... doesn't seem ideal :slightly_smiling_face:. But we can at least follow the recommendation of separating the various bits rather than using 14-digit numbers. (Also, parsers might choke on those, since they don't fit in 32-bits, and version numbers don't usually contain components larger than 32-bit numbers.) Also, I believe hyphens are not allowed at all; plus is the correct separator to allow a "Local version identifiers".
</blockquote></details>

(2) In that case, the current tar.gz packaging builds will need to change to use the more specific PEP format, so that Drake is consistent across the board.  That doesn't need to hold up _this_ PR, though.   When that change happens, ideally we would have only one copy of this ephemeral-version-creation code.  Until that fix happens, the two cmake files will need to each have their own fork of this logic.

jwnimmer-tri (dismissed):
> Reviewed 5 of 6 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

mwoehlke-kitware (commented):
Neither of those appears to have actually built the new configuration; was the intent only to test that this didn't break existing builds?

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-:-Mpr4l0j2ENh6ALLzLiQ:bhh3lwe)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

BetsyMcPhail (commented):
Yes, just a sanity check before merging.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-:-MprGo5fDRd6vDeYkyQR:b-72ozpv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

mwoehlke-kitware (commented):
Looks like CI is working now. Everyone okay with the most recent version?

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-:-Mq0ZMjwCH4bbjjR-VLW:bfqevjw)* status: 5 of 8 files reviewed, all discussions resolved (waiting on @jwnimmer-tri)

BetsyMcPhail (commented):
Passing build: https://drake-jenkins.csail.mit.edu/view/Wheel/job/linux-focal-unprovisioned-gcc-wheel-experimental-release/5/

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-:-Mq0bHY7-tk6WTGtzQAI:b5kq5x8)* status: 5 of 8 files reviewed, all discussions resolved (waiting on @jwnimmer-tri)

BetsyMcPhail (approved):
:lgtm:

> Reviewed 5 of 6 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-:-Mq0mLfn2yeg0_QFSj3Y:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

mwoehlke-kitware:
@jwnimmer-tri, any objections?

jwnimmer-tri (changes requested):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129)* status: all files reviewed, 1 unresolved discussion (waiting on @mwoehlke-kitware)

___
*[driver/configurations/wheel/step-provision.cmake, line 40 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-MqFh0Qi-cz4nddUisd7:-MqFh0Qj6ytL6z-lmxsb:b6k81nd) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/1d7ffc81fa6a1ad8e5df67c82e1cbada0f44ddc3/driver/configurations/wheel/step-provision.cmake#L40)):*
> ```CMake
>   RESULT_VARIABLE DOCKER_RESULT_VARIABLE)
> if(DOCKER_RESULT_VARIABLE EQUAL 0)
>   return()
> ```

I haven't dug in to all of the call sequences through these files, but I wonder if this logic is correct.

If a Jenkins build is named "unprovisioned", then all of the provisioning code should always run 100% of the time, without any short-circuits.

If the Jenkins build is not named "unprovisioned", then no provisioning code should ever be run.

The most recent push here seems to be splitting the difference, which is not correct?

mwoehlke-kitware:
___
*[driver/configurations/wheel/step-provision.cmake, line 40 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-MqFh0Qi-cz4nddUisd7:-MqFkXJk2SUxDIgLoDlK:b884p25) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/1d7ffc81fa6a1ad8e5df67c82e1cbada0f44ddc3/driver/configurations/wheel/step-provision.cmake#L40)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

I haven't dug in to all of the call sequences through these files, but I wonder if this logic is correct.

If a Jenkins build is named "unprovisioned", then all of the provisioning code should always run 100% of the time, without any short-circuits.

If the Jenkins build is not named "unprovisioned", then no provisioning code should ever be run.

The most recent push here seems to be splitting the difference, which is not correct?
</blockquote></details>

"Provisioned" images don't have Docker, nor does provisioning for a "normal" Drake build install Docker. This skip is actually so that `run_local.sh` doesn't try to `sudo` the script to install Docker. (Note that "publish" configurations are also subject to this; I guess no one is trying to run those via `run_local.sh`.)

An alternative approach would be to install Docker on all of the provisioned images and modify the provisioning so that unprovisioned "wheel" or "publish" configurations (if we even want those¹) also install Docker. (¹If not, we should probably consider removing `setup/docker/install_prereqs`.)

jwnimmer-tri (changes requested):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129)* status: all files reviewed, 1 unresolved discussion (waiting on @mwoehlke-kitware)

___
*[driver/configurations/wheel/step-provision.cmake, line 40 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-MqFh0Qi-cz4nddUisd7:-MqFnjbd2AsLydeLcYum:b-mj1x09) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/1d7ffc81fa6a1ad8e5df67c82e1cbada0f44ddc3/driver/configurations/wheel/step-provision.cmake#L40)):*
> An alternative approach would be to install Docker on all of the provisioned images ...

I don't need provisioned images to have Docker installed.  I need this code to fail-fast if it is attempted to be run on a build without "unprovisioned" in its name.  Does it already do that?

> This skip is actually so that run_local.sh doesn't try to sudo the script to install Docker.

Ah, okay.  I always forget about that script.  I think you're the only person who uses it.

In that case, the early-return here makes sense, but it should have a comment that the short-circuit is intended only to occur for run_local users who already have docker installed.

mwoehlke-kitware:
___
*[driver/configurations/wheel/step-provision.cmake, line 40 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-MqFh0Qi-cz4nddUisd7:-MqFoXDK36RyfLMhYxiL:b-p0ck56) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/1d7ffc81fa6a1ad8e5df67c82e1cbada0f44ddc3/driver/configurations/wheel/step-provision.cmake#L40)):*
> I need this code to fail-fast if it is attempted to be run on a build without "unprovisioned" in its name.  Does it already do that?

Uh... no? Why does it need to do that? This should run fine on either provisioned or unprovisioned images. Both will need to install Docker (just as "publish" configurations do).

>  I think you're the only person who uses [`run_local.sh`].

Well, it's super-helpful for development, to not have to spin up an AWS instance during early testing :slightly_smiling_face:.

jwnimmer-tri (approved):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/129)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

___
*[driver/configurations/wheel/step-provision.cmake, line 40 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ci/129#-MqFh0Qi-cz4nddUisd7:-MqFpemw2lDL8-XjQMFy:b-97u6p8) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/1d7ffc81fa6a1ad8e5df67c82e1cbada0f44ddc3/driver/configurations/wheel/step-provision.cmake#L40)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

> I need this code to fail-fast if it is attempted to be run on a build without "unprovisioned" in its name.  Does it already do that?

Uh... no? Why does it need to do that? This should run fine on either provisioned or unprovisioned images. Both will need to install Docker (just as "publish" configurations do).

>  I think you're the only person who uses [`run_local.sh`].

Well, it's super-helpful for development, to not have to spin up an AWS instance during early testing :slightly_smiling_face:.
</blockquote></details>

Whatever, this is a waste of my time.  Go ahead and merge.

PR #153, opened 2022-05-04
Use `--config=` for consistency

mwoehlke-kitware:
Use `--config=${COMPILER}` instead of `--config ${COMPILER}` for consistency with documentation and with other uses of `--config`. This is purely an aesthetic change; Bazel should parse the arguments the same in either case.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/153)

### DIFF
--- a/driver/configurations/bazel.cmake
+++ b/driver/configurations/bazel.cmake
@@ -117,7 +117,7 @@ if(REMOTE_CACHE)
   )
 endif()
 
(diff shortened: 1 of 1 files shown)


jwnimmer-tri:
:lgtm:

jwnimmer-tri (approved):

PR #162, opened 2022-06-08
Don't clean up macOS wheel build

mwoehlke-kitware:
Don't try to clean up after the macOS wheel build, because the wheel gets dumped into one of the trees that "cleaning up" would try to clean up. This would result in the wheel being deleted, except that the builder can't delete it due to permissions, which would be flagged as a build error.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/162)

### DIFF
--- a/driver/configurations/wheel.cmake
+++ b/driver/configurations/wheel.cmake
@@ -104,6 +104,12 @@ else()
 endif()
 
 if(APPLE)
+  # Don't try to clean up after the build, because the wheel gets dumped into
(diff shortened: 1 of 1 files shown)


mwoehlke-kitware:
Test build here: https://drake-jenkins.csail.mit.edu/job/mac-monterey-unprovisioned-clang-wheel-experimental-snopt-mosek-release/7/

mwoehlke-kitware:
See #163 instead.

PR #164, opened 2022-06-16
Use a Python script to upload to AWS

mwoehlke-kitware:
Replace the nasty CMake logic used to upload artifacts (currently, just the binary package, but we need to add other artifacts in the near future) with a Python script. This should be more flexible, and the code should be much easier to read and understand. A small CMake helper function for invoking the script has also been added.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164)


mwoehlke-kitware:
Test build (I hope) at https://drake-jenkins.csail.mit.edu/view/Packaging/job/linux-focal-unprovisioned-gcc-bazel-experimental-snopt-mosek-packaging/.

We still need to verify if the expirations are acceptable. I went ~50% longer than one week / one month.

mwoehlke-kitware:
Okay, working (AFAICT) build here: https://drake-jenkins.csail.mit.edu/view/Packaging/job/linux-focal-unprovisioned-gcc-bazel-experimental-snopt-mosek-packaging/19/console.

Note that the output has changed a bit; not sure if that's going to cause other problems...

mwoehlke-kitware:
Sample 'before' output:
```
[0:29:02.393]  -- Uploading package archive 1 of 1 to AWS S3...
[0:29:02.393]  '/usr/bin/aws' 's3' 'cp' '--acl' 'public-read' '--cache-control' 'max-age=2419200' '--storage-class' 'STANDARD' '/media/ephemeral0/ubuntu/workspace/linux-focal-unprovisioned-gcc-bazel-experimental-snopt-mosek-packaging/drake-20220609160939-88625e5dbfc7eae7684146568562631abf2857c7-focal.tar.gz' 's3://drake-packages/drake/experimental/drake-20220609160939-88625e5dbfc7eae7684146568562631abf2857c7-focal.tar.gz'
[0:29:03.471]  Completed 256.0 KiB/232.4 MiB (2.5 MiB/s) with 1 file(s) remaining
Completed 512.0 KiB/232.4 MiB (4.9 MiB/s) with 1 file(s) remaining
...many repeats...
Completed 232.4 MiB/232.4 MiB (175.8 MiB/s) with 1 file(s) remaining
upload: ./drake-20220609160939-88625e5dbfc7eae7684146568562631abf2857c7-focal.tar.gz to s3://drake-packages/drake/experimental/drake-20220609160939-88625e5dbfc7eae7684146568562631abf2857c7-focal.tar.gz
[0:29:03.471]  -- Package URL 1 of 1: https://drake-packages.csail.mit.edu/drake/experimental/drake-20220609160939-88625e5dbfc7eae7684146568562631abf2857c7-focal.tar.gz
[0:29:04.031]  -- Uploading package archive checksum 1 of 1 to AWS S3...
[0:29:04.031]  '/usr/bin/aws' 's3' 'cp' '--acl' 'public-read' '--cache-control' 'max-age=2419200' '--storage-class' 'STANDARD' '/media/ephemeral0/ubuntu/workspace/linux-focal-unprovisioned-gcc-bazel-experimental-snopt-mosek-packaging/drake-20220609160939-88625e5dbfc7eae7684146568562631abf2857c7-focal.tar.gz.sha512' 's3://drake-packages/drake/experimental/drake-20220609160939-88625e5dbfc7eae7684146568562631abf2857c7-focal.tar.gz.sha512'
[0:29:04.287]  Completed 204 Bytes/204 Bytes (2.8 KiB/s) with 1 file(s) remaining
upload: ./drake-20220609160939-88625e5dbfc7eae7684146568562631abf2857c7-focal.tar.gz.sha512 to s3://drake-packages/drake/experimental/drake-20220609160939-88625e5dbfc7eae7684146568562631abf2857c7-focal.tar.gz.sha512
```

Sample 'after' output:
```
[0:19:34.999]  -- Uploading drake-20220616204833-57704aae6b770cb7d78a3cb1a723f6f60b5bc263-focal.tar.gz to AWS S3...
[0:19:34.999]  ['/usr/bin/aws', 's3', 'cp', '--acl', 'public-read', '--storage-class', 'STANDARD', '--cache-control', 'max-age=864000', '/media/ephemeral0/ubuntu/workspace/linux-focal-unprovisioned-gcc-bazel-experimental-snopt-mosek-packaging/drake-20220616204833-57704aae6b770cb7d78a3cb1a723f6f60b5bc263-focal.tar.gz', 's3://drake-packages/drake/experimental/drake-20220616204833-57704aae6b770cb7d78a3cb1a723f6f60b5bc263-focal.tar.gz']
[0:19:36.851]  Completed 256.0 KiB/233.1 MiB (2.7 MiB/s) with 1 file(s) remaining
Completed 512.0 KiB/233.1 MiB (5.3 MiB/s) with 1 file(s) remaining
...many repeats...
Completed 233.1 MiB/233.1 MiB (199.1 MiB/s) with 1 file(s) remaining
upload: ./drake-20220616204833-57704aae6b770cb7d78a3cb1a723f6f60b5bc263-focal.tar.gz to s3://drake-packages/drake/experimental/drake-20220616204833-57704aae6b770cb7d78a3cb1a723f6f60b5bc263-focal.tar.gz
[0:19:36.851]  -- Upload complete: https://drake-packages.csail.mit.edu/drake/experimental/drake-20220616204833-57704aae6b770cb7d78a3cb1a723f6f60b5bc263-focal.tar.gz
[0:19:37.414]  -- Uploading drake-20220616204833-57704aae6b770cb7d78a3cb1a723f6f60b5bc263-focal.tar.gz checksum to AWS S3...
[0:19:37.414]  ['/usr/bin/aws', 's3', 'cp', '--acl', 'public-read', '--storage-class', 'STANDARD', '--cache-control', 'max-age=864000', '/tmp/tmpoj0wo01l', 's3://drake-packages/drake/experimental/drake-20220616204833-57704aae6b770cb7d78a3cb1a723f6f60b5bc263-focal.tar.gz.sha512']
[0:19:37.670]  Completed 205 Bytes/205 Bytes (3.0 KiB/s) with 1 file(s) remaining
upload: ../../../../../tmp/tmpoj0wo01l to s3://drake-packages/drake/experimental/drake-20220616204833-57704aae6b770cb7d78a3cb1a723f6f60b5bc263-focal.tar.gz.sha512
```

svenevs (changes requested):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164)* status: 0 of 5 files reviewed, 17 unresolved discussions (waiting on @mwoehlke-kitware and @svenevs)

___
*[`driver/configurations/cache.cmake` line 410 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4igrMl-8BVL2fdQRb3:-N4igrMl-8BVL2fdQRb4:b-81nte3) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/driver/configurations/cache.cmake#L410)):*
> ```CMake
> 
> if(REMOTE_CACHE)
>   get_filename_component(PYTHON_EXECUTABLE "${DASHBOARD_PYTHON_COMMAND}" REALPATH)
> ```

Do we have any idea what jobs are affected by these changes?  THey seem fine but it would be nice to confirm.
___
*[`tools/upload-to-aws.py` line 6 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iRIFACE77QLeKty0G:-N4iRIFACE77QLeKty0H:bfq4znz) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L6)):*
> ```Python
> #
> # Copyright (c) 2019, Massachusetts Institute of Technology.
> # Copyright (c) 2019, Toyota Research Institute.
> ```

It's 2022 :wink:
___
*[`tools/upload-to-aws.py` line 49 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ibiyjBPLWWnuUSe1h:-N4ibiyjBPLWWnuUSe1i:b-ks64rf) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L49)):*
> ```Python
> archive_storage_class = 'STANDARD'
> max_attempts = 3
> backoff_delay = 15
> ```

I think these variables should just be local to the `upload` method.
___
*[`tools/upload-to-aws.py` line 52 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iZKYq7InYKroPJJJ0:-N4iZKYq7InYKroPJJJ1:b-fdrfsg) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L52)):*
> ```Python
> 
> 
> def age(days=0, hours=0, minutes=0):
> ```

Minor, I think it would be better to rename this to `total_seconds` and use stdlib

```py
from datetime import timedelta

def total_seconds(days=0, hours=0, minutes=0):
    """Return the integral number of seconds for the specified time duration."""
    return int(timedelta(days=days, hours=hours, minutes=minutes).total_seconds())
```

Current implementation will have issues if anyone ever does like `hours=1.5`, s3 needs ints AFAICT.

That said, an alternative would be to in `max_age` just do something like

```py
def max_age(options, latest):
    if options.nightly:
        kwargs = {"hours": 18} if latest else {"days": 45}
    else:
        kwargs = {"minutes": 30} if latest else {"days": 10}
    return int(timedelta(**kwargs).total_seconds())
```

and delete the function altogether.  Depends on how you feel about kwargs (I'm a fan), main point is make sure return is an `int`.
___
*[`tools/upload-to-aws.py` line 59 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ieEzZ7mvJbOs-Z94L:-N4ieEz_1ooISSQhmhWO:b-jlh1gw) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L59)):*
> ```Python
> 
> 
> def max_age(options, latest=False):
> ```

I think it would be better to require `latest` be keyword only without a default value.

```py
def max_age(options, *, latest: bool):
    # ...
```

caller always required to provide `latest=True|False` as a result, keeps things easier to follow.
___
*[`tools/upload-to-aws.py` line 61 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iXZdR2E8tuKymA7U4:-N4iXZdR2E8tuKymA7U5:brzly2r) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L61)):*
> ```Python
> def max_age(options, latest=False):
>     """
>     Calculates the appropriate expiration for an artifact.
> ```

The docstring should use explicit verbiage explaining what is being done here for each track.  Explain that it is going to return an integer for time in seconds.  For now we're mirroring this:

https://github.com/RobotLocomotion/drake-ci/blob/8474a8e7bc72e6591d4e85811a11e5360b65364c/driver/configurations/bazel/step-upload-package-archive.cmake#L40-L46

In the future we are to change it but I need to confirm the desired duration.
___
*[`tools/upload-to-aws.py` line 69 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ia3Wg9nGf4zn36e92:-N4ia3Wg9nGf4zn36e93:b-vl6wd6) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L69)):*
> ```Python
> 
> 
> def upload(path, name, expiration, options, noisy=False):
> ```

I think we should just remove `noisy` as a parameter and always print things.  There's no place where `noisy=False` is used and the more we have logged the better IMO, particularly for the final download URI, you always want that in the logs.
___
*[`tools/upload-to-aws.py` line 85 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4idNhi8id6CrgZeAaK:-N4idNhi8id6CrgZeAaL:b-mp5m0n) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L85)):*
> ```Python
>         print(f'-- Uploading {name} to AWS S3...', flush=True)
> 
>     print(command, file=sys.stderr, flush=True)
> ```

Why does this one go to `stderr`?  It might be nice to print it slightly more formatted, something like `print(f"-- Executing: {' '.join(command)}", ...)`
___
*[`tools/upload-to-aws.py` line 98 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ievet1CKQ_jk2o6l9:-N4ievet1CKQ_jk2o6lA:b-3x89sa) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L98)):*
> ```Python
>                 sleep(backoff_delay)
>             else:
>                 raise
> ```

Rather than a bare `raise` without any explanation can we print an error message ("attempted X uploads but all were unsuccessful" or whatever) and then `sys.exit(1)`
___
*[`tools/upload-to-aws.py` line 119 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ig0aL0kn4hflNyPsB:-N4ig0aL0kn4hflNyPsC:b-ii4dca) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L119)):*
> ```Python
>             checksum.update(data)
> 
>     with tempfile.NamedTemporaryFile(mode='w') as f:
> ```

I don't think it's a good idea to be uploading the file while it is still open, we should just create `f"{path}.sha512"` as is done currently.
___
*[`tools/upload-to-aws.py` line 140 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iX-ZN5LCRzWjjQGkS:-N4iX-ZN5LCRzWjjQGkT:bog9lyo) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L140)):*
> ```Python
> 
>     if not options.experimental:
>         m = re.match(r'^drake-[^-]+(-[0-9a-f]{40})?-(.*)$', name)
> ```

Please add a comment explaining (using examples to illustrate of existing archive names) what this regex is and why `group(2)` is needed.  Nobody looking at this in the future is going to have any idea what this is.
___
*[`tools/upload-to-aws.py` line 164 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iTE_e5z7_R7-Jzo8W:-N4iTE_e5z7_R7-Jzo8X:b-i1mx8i) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L164)):*
> ```Python
>     parser.add_argument(
>         '--aws', type=str, default='aws',
>         help='Path to AWS executable')
> ```

We can keep this as a parameter, but should assert that it is an executable path that exists.  Alternatively, rather than passing it as a parameter we can just use [`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which) and assert that the return is `not None`.
___
*[`tools/upload-to-aws.py` line 166 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iT0cZ1vKZyJwIenjH:-N4iT0cZ1vKZyJwIenjI:bphpb5g) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L166)):*
> ```Python
>         help='Path to AWS executable')
>     parser.add_argument(
>         '--track', type=str, required=True,
> ```

We should add `choices={"Continuous", "Nightly", "Experimental"},` so that if there's ever something new created this will crash and we know to update logic here.  Per other discussion, just make a constant `track_choices = {"Continuous", "Nightly", "Experimental"}`
___
*[`tools/upload-to-aws.py` line 169 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iU3jK-jbOHa6rx1H6:-N4iU3jK-jbOHa6rx1H7:b-19aebz) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L169)):*
> ```Python
>         help='CI track of artifact')
> 
>     options = parser.parse_args(args)
> ```

I think it would be more preferable to behave like the other scripts in drake, it's a little strange to pass in `args` here.  Suggestion:

```py
def main():
    parser = ArgumentParser()
    # ... add args ...
    args = parser.parse_args()
```

I guess we can keep it as `options` but it's fairly standard to be called `args`.
___
*[`tools/upload-to-aws.py` line 170 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iWohl96LVVPIjr3tz:-N4iWohmAvbJ-mVSqFLe:bt9ke80) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L170)):*
> ```Python
> 
>     options = parser.parse_args(args)
> ```

We should assert that the input `artifact` is a file, if not `parser.error`
___
*[`tools/upload-to-aws.py` line 172 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iVVUOCF-TrlW1wwEP:-N4iVVUOCF-TrlW1wwEQ:bw2x4q1) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L172)):*
> ```Python
> 
>     options.track = options.track.lower()
>     set_track_option(options, 'nightly')
> ```

It's not clear to me why having a helper function wrapping it is preferred, it's a little harder to follow in my opinion.  I think the following would be easier to follow keeping it all in one place:

```py
options.track = options.track.lower()

# Enable helper methods to be able to query e.g., `if options.nightly` rather than needing
# to check `if options.track == "nightly"`.
for track in track_choices:
    track_lower = track.lower()
    setattr(options, track_lower, options.track == track_lower)
```

Initially reading the three separate `set_track_option` would make me think that more than just a boolean is being added (e.g., age or something).
___
*[`tools/upload-to-aws.py` line 176 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4igHS945Hi0R3UUYAj:-N4igHS945Hi0R3UUYAk:b-rmoltk) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L176)):*
> ```Python
>     set_track_option(options, 'experimental')
> 
>     upload_artifacts(options)
> ```

We should wrap this whole process in a `try-except` catch all in the event that something undetected goes wrong, and just print the error message and set the exit code.  Unhandled exceptions are ugly :wink:

mwoehlke-kitware (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164)* status: 0 of 5 files reviewed, 16 unresolved discussions (waiting on @mwoehlke-kitware and @svenevs)

___
*[`driver/configurations/cache.cmake` line 410 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4igrMl-8BVL2fdQRb3:-N4m00b-7OyeSPRzWqz0:b-rzo7h0) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/driver/configurations/cache.cmake#L410)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

Do we have any idea what jobs are affected by these changes?  THey seem fine but it would be nice to confirm.
</blockquote></details>

Not really. As long as at least one caching job still works, though, it's probably fine.
___
*[`tools/upload-to-aws.py` line 6 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iRIFACE77QLeKty0G:-N4m0qQu7ipI4IaVxxCV:b-896fix) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L6)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

It's 2022 :wink:
</blockquote></details>

Done.
___
*[`tools/upload-to-aws.py` line 49 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ibiyjBPLWWnuUSe1h:-N4m0f9f7jsC17iAw1N8:b-kdfzhq) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L49)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

I think these variables should just be local to the `upload` method.
</blockquote></details>

If they're local, there would be little point in them being variables at all. As globals, they're easy to find and it's easy to see they affect the script operation.
___
*[`tools/upload-to-aws.py` line 52 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iZKYq7InYKroPJJJ0:-N4m1p9R4pHkWLPALe6J:b-w7k5be) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L52)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

Minor, I think it would be better to rename this to `total_seconds` and use stdlib

```py
from datetime import timedelta

def total_seconds(days=0, hours=0, minutes=0):
    """Return the integral number of seconds for the specified time duration."""
    return int(timedelta(days=days, hours=hours, minutes=minutes).total_seconds())
```

Current implementation will have issues if anyone ever does like `hours=1.5`, s3 needs ints AFAICT.

That said, an alternative would be to in `max_age` just do something like

```py
def max_age(options, latest):
    if options.nightly:
        kwargs = {"hours": 18} if latest else {"days": 45}
    else:
        kwargs = {"minutes": 30} if latest else {"days": 10}
    return int(timedelta(**kwargs).total_seconds())
```

and delete the function altogether.  Depends on how you feel about kwargs (I'm a fan), main point is make sure return is an `int`.


</blockquote></details>

Done. (I took a somewhat in-between approach.) Thanks for pointing out `timedelta`; I was wondering if there was something like that, but didn't spot it readily.
___
*[`tools/upload-to-aws.py` line 59 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ieEzZ7mvJbOs-Z94L:-N4m2lo4C8tz2TOtyUYD:b-896fix) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L59)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

I think it would be better to require `latest` be keyword only without a default value.

```py
def max_age(options, *, latest: bool):
    # ...
```

caller always required to provide `latest=True|False` as a result, keeps things easier to follow.
</blockquote></details>

Done.
___
*[`tools/upload-to-aws.py` line 61 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iXZdR2E8tuKymA7U4:-N4m37AeDDA9CF12KP8d:b-4lascv) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L61)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

The docstring should use explicit verbiage explaining what is being done here for each track.  Explain that it is going to return an integer for time in seconds.  For now we're mirroring this:

https://github.com/RobotLocomotion/drake-ci/blob/8474a8e7bc72e6591d4e85811a11e5360b65364c/driver/configurations/bazel/step-upload-package-archive.cmake#L40-L46

In the future we are to change it but I need to confirm the desired duration.
</blockquote></details>

I don't follow.
___
*[`tools/upload-to-aws.py` line 69 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ia3Wg9nGf4zn36e92:-N4m39mtDomWltnk9PoA:b-w0owk2) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L69)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

I think we should just remove `noisy` as a parameter and always print things.  There's no place where `noisy=False` is used and the more we have logged the better IMO, particularly for the final download URI, you always want that in the logs.
</blockquote></details>

Checksums are half as noisy, and use a different notice.
___
*[`tools/upload-to-aws.py` line 85 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4idNhi8id6CrgZeAaK:-N4m3LFv4hChevu807in:b9l52bn) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L85)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

Why does this one go to `stderr`?  It might be nice to print it slightly more formatted, something like `print(f"-- Executing: {' '.join(command)}", ...)`
</blockquote></details>

...because that's what CMake did? Whether it makes sense or not is debatable.

I **do not** want to use `'  '.join(command)` because that loses information, even compared to CMake (which doesn't print `[]`s or I believe `,`s, but *does* quote each argument). "Pretty" is not a goal here. Precision *is*.
___
*[`tools/upload-to-aws.py` line 98 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ievet1CKQ_jk2o6l9:-N4m3vqRASZwzn1fdovU:b-n83k8i) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L98)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

Rather than a bare `raise` without any explanation can we print an error message ("attempted X uploads but all were unsuccessful" or whatever) and then `sys.exit(1)`
</blockquote></details>

A bare `raise` is a re-throw; that's deliberate. It should be obvious from the output that attempts were made. If we change it, I would want to dump *every* exception rather than only the last.
___
*[`tools/upload-to-aws.py` line 169 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iU3jK-jbOHa6rx1H6:-N4m4zfW9TL6dTmtRoGi:b-bafz5s) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L169)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

I think it would be more preferable to behave like the other scripts in drake, it's a little strange to pass in `args` here.  Suggestion:

```py
def main():
    parser = ArgumentParser()
    # ... add args ...
    args = parser.parse_args()
```

I guess we can keep it as `options` but it's fairly standard to be called `args`.
</blockquote></details>

That's how I *always* write Python scripts, and it's consistent with the wheel scripts. It's also more flexible (can pass something other than `sys.argv[1:]` as arguments) and more correct; the result of parsing is *not* the arguments themselves.
___
*[`tools/upload-to-aws.py` line 176 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4igHS945Hi0R3UUYAj:-N4m5_co5y_KS8la6vwW:b-vihb07) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L176)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

We should wrap this whole process in a `try-except` catch all in the event that something undetected goes wrong, and just print the error message and set the exit code.  Unhandled exceptions are ugly :wink:
</blockquote></details>

So... we should implement an exception handler that exactly duplicates the default exception handler? Why?

mwoehlke-kitware:
Updated CI at https://drake-jenkins.csail.mit.edu/view/Packaging/job/linux-focal-unprovisioned-gcc-bazel-experimental-snopt-mosek-packaging/20.

Artifacts:
- https://drake-packages.csail.mit.edu/drake/experimental/drake-20220617162838-5c86fd443b7cbc07ac13411373a73df74648ec46-focal.tar.gz
- https://drake-packages.csail.mit.edu/drake/experimental/drake-20220617162838-5c86fd443b7cbc07ac13411373a73df74648ec46-focal.tar.gz.sha512
```
9f90ba627214d5d9e8314a8e772ffd910949da0f9b1561b47b2d32cdc892f32ce3c2382d08bf2214f432fda79e6099f90213909b61ac34ffe7a9cbb7c474e010  drake-20220617162838-5c86fd443b7cbc07ac13411373a73df74648ec46-focal.tar.gz
```

svenevs (changes requested):
> Reviewed 2 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164)* status: 2 of 5 files reviewed, 6 unresolved discussions (waiting on @mwoehlke-kitware and @svenevs)

___
*[`tools/upload-to-aws.py` line 49 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ibiyjBPLWWnuUSe1h:-N4mUTYb2__rAEYtOD8T:b-r13bew) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L49)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

If they're local, there would be little point in them being variables at all. As globals, they're easy to find and it's easy to see they affect the script operation.
</blockquote></details>

OK, I don't have a strong opinion.  It's easier to follow if they're declared closer to where they are used
___
*[`tools/upload-to-aws.py` line 52 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iZKYq7InYKroPJJJ0:-N4mWr1M5UNU50OnpVMi:b43i9yb) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L52)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

Done. (I took a somewhat in-between approach.) Thanks for pointing out `timedelta`; I was wondering if there was something like that, but didn't spot it readily.
</blockquote></details>

In-between looks good, this function can be deleted now right?
___
*[`tools/upload-to-aws.py` line 59 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ieEzZ7mvJbOs-Z94L:-N4mWYsf-OmE2z7XtJ-D:b-5tgt9l) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L59)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Missing the `, *,` which makes it keyword only.  It's so that when somebody does calls `max_age(options, False)` it's an error, making them do `max_age(options, latest=False)` -- keeps it obvious what the boolean represents at the calling site.
___
*[`tools/upload-to-aws.py` line 61 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iXZdR2E8tuKymA7U4:-N4maAzV-CC0NvPQ5VnD:b-4e0kyl) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L61)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

I don't follow.
</blockquote></details>

OK, not needed anymore given changes -- it's much clearer now that seconds are returned
___
*[`tools/upload-to-aws.py` line 69 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ia3Wg9nGf4zn36e92:-N4mX6OJ6ARFIuiT38U4:botfgxh) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L69)):*
> Checksums are half as noisy

If we're going to keep it, I'd prefer it be like `max_age` and require `options, *, noisy: bool`.  Since during review it wasn't clear that a `noisy=False` was happening (would need to update call in `upload_checksum` to provide `noisy=False`.

But the question is why are they different?

> particularly for the final download URI, you always want that in the logs

Same is true for the uploaded sha file which currently does not happen.  This is the main point of interest here.  Dev convenience.

Per tempfile discussion, if the contention is just the different first notice, provide `suffix='.sha512'` to tempfile creation?  Or move the first notice out of the method?  I only care that the final uploaded file for me the person reading the logs is available.  Dedent that from `noisy`?  However you want to do it.  It's not a very clear parameter name.
___
*[`tools/upload-to-aws.py` line 85 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4idNhi8id6CrgZeAaK:-N4mVzIb5SQm22nlu1Vt:b5lqzmm) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L85)):*
> Whether it makes sense or not is debatable.

It doesn't, lets remove `stderr`, nothing else uses it.  We aren't beholden to the arbitrary decisions of old.

>  "Pretty" is not a goal here. Precision is.

Good point, that may help debug it one day.
___
*[`tools/upload-to-aws.py` line 98 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ievet1CKQ_jk2o6l9:-N4mUrBV2DfHO6ZMhzjV:bxe2gsg) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L98)):*
> It should be obvious from the output that attempts were made

It won't be immediately obvious why the script stopped.  Something like `print(f"ERROR: attempted {max_attempts} unsuccessful uploads.")`.  Makes it searchable, `ERROR` i think makes it show up in cdash.  The `CalledProcessError` tracebacks are somewhat useless, but I'm more interested in a user facing (person searching the logs) error message explaining what went wrong.  keeping `raise` is fine
___
*[`tools/upload-to-aws.py` line 119 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ig0aL0kn4hflNyPsB:-N4mYGjH0iXAZe2dXX3U:biomr3u) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L119)):*
> That requires that the artifact's containing directory is writable ... any desire to make this more general

There isn't, we know with certainty we can write to the directory where the artifact came from in CI.

> I don't see why it's problematic.

The file has not been closed yet when you ask another process to start reading it, that isn't always successful and if it does fail it will be hard to understand what went wrong from CI logs.  Alternative:

```py
fd, path = tempfile.mkstemp(text=True)
os.write(fd, f'{checksum.hexdigest()}  {name}\n')
os.close(fd)
upload(...)
os.remove(path)
```
___
*[`tools/upload-to-aws.py` line 164 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iTE_e5z7_R7-Jzo8W:-N4mZyxrE4i2qN3KOGCl:b-mnu2lh) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L164)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

I think that will be sufficiently obvious when `check_call` can't execute it. Otherwise we need additional logic to determine if a name with no path is executable. This isn't a user-facing tool; I question how much effort should be devoted to making it user-friendly.
</blockquote></details>

OK, cmake side validates it.
___
*[`tools/upload-to-aws.py` line 166 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iT0cZ1vKZyJwIenjH:-N4m_3CB4BGl_U9ZwGjh:b-r8959k) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L166)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

LGTM, `type=str.lower` is very cool :sunglasses: never seen that!
___
*[`tools/upload-to-aws.py` line 169 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iU3jK-jbOHa6rx1H6:-N4m_CqU9RfJHt4c69Un:bsiezgw) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L169)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

That's how I *always* write Python scripts, and it's consistent with the wheel scripts. It's also more flexible (can pass something other than `sys.argv[1:]` as arguments) and more correct; the result of parsing is *not* the arguments themselves.
</blockquote></details>

OK, nowhere else uses that but it's not a deal breaker for me.
___
*[`tools/upload-to-aws.py` line 176 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4igHS945Hi0R3UUYAj:-N4m_M48DxiT-h0LR3Zl:baoqq1l) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L176)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

So... we should implement an exception handler that exactly duplicates the default exception handler? Why?
</blockquote></details>

OK, no reason in particular, other edits were being requested so pointed it out.

svenevs (changes requested):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164)* status: 2 of 5 files reviewed, 6 unresolved discussions (waiting on @mwoehlke-kitware and @svenevs)

___
*[`driver/configurations/cache.cmake` line 410 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4igrMl-8BVL2fdQRb3:-N4mcKdtCO7V8PL_qTgb:b-j1mm19) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/driver/configurations/cache.cmake#L410)):*
<details><summary><i>Previously, mwoehlke-kitware (Matthew Woehlke) wrote…</i></summary><blockquote>

Not really. As long as at least one caching job still works, though, it's probably fine.
</blockquote></details>

this job should use caching https://drake-jenkins.csail.mit.edu/job/linux-focal-gcc-bazel-experimental-release/11086/

jwnimmer-tri:
*[`tools/upload-to-aws.py` line 49 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ibiyjBPLWWnuUSe1h:-N4mcMAK9Cbogw5_OLWr:b3w5jbf) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L49)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

OK, I don't have a strong opinion.  It's easier to follow if they're declared closer to where they are used
</blockquote></details>

It's fine to have module-level constants, but in that case they should follow the pep-008 CONSTANT_NAME convention so it's clear at their point-of-use that they are readonly.

https://peps.python.org/pep-0008/#constants

svenevs (commented):
> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164)* status: 3 of 5 files reviewed, 5 unresolved discussions (waiting on @jwnimmer-tri, @mwoehlke-kitware, and @svenevs)

___
*[`driver/configurations/cache.cmake` line 410 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4igrMl-8BVL2fdQRb3:-N4mdnvNAATh39spZq3T:bffxadt) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/driver/configurations/cache.cmake#L410)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

this job should use caching https://drake-jenkins.csail.mit.edu/job/linux-focal-gcc-bazel-experimental-release/11086/
</blockquote></details>

OK cache was used

mwoehlke-kitware (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164)* status: 3 of 5 files reviewed, 5 unresolved discussions (waiting on @jwnimmer-tri and @svenevs)

___
*[`tools/upload-to-aws.py` line 49 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ibiyjBPLWWnuUSe1h:-N4mrfLBBDM6-n5NfHiz:bedemrd) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L49)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

^^ to avoid multiple push, depending on decision make sure docstring for `upload` referencing `max_attempts` is updated accordingly
</blockquote></details>

Done. (Global find/replace found that, anyway :slightly_smiling_face:.)
___
*[`tools/upload-to-aws.py` line 52 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iZKYq7InYKroPJJJ0:-N4mrnSv-c6GQC4OO5_r:bc721zd) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L52)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

In-between looks good, this function can be deleted now right?
</blockquote></details>

Done. (Yup, missed it in the previous update. Caught it before seeing this comment, but thanks for keeping me honest, anyway!)
___
*[`tools/upload-to-aws.py` line 59 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ieEzZ7mvJbOs-Z94L:-N4mrAMv3FoxBIGpudQq:b-896fix) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L59)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

Missing the `, *,` which makes it keyword only.  It's so that when somebody does calls `max_age(options, False)` it's an error, making them do `max_age(options, latest=False)` -- keeps it obvious what the boolean represents at the calling site.
</blockquote></details>

Done.
___
*[`tools/upload-to-aws.py` line 61 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iXZdR2E8tuKymA7U4:-N4moRhCBo75HJODxo05:bfe99mk) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L61)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

OK, not needed anymore given changes -- it's much clearer now that seconds are returned
</blockquote></details>

Ah, okay, I think I got hung up on your first sentence and missed the second :flushed:.
___
*[`tools/upload-to-aws.py` line 69 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ia3Wg9nGf4zn36e92:-N4mrBow50QW-lRSltz0:b-qbvcu3) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L69)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

> Checksums are half as noisy

If we're going to keep it, I'd prefer it be like `max_age` and require `options, *, noisy: bool`.  Since during review it wasn't clear that a `noisy=False` was happening (would need to update call in `upload_checksum` to provide `noisy=False`.

But the question is why are they different?

> particularly for the final download URI, you always want that in the logs

Same is true for the uploaded sha file which currently does not happen.  This is the main point of interest here.  Dev convenience.

Per tempfile discussion, if the contention is just the different first notice, provide `suffix='.sha512'` to tempfile creation?  Or move the first notice out of the method?  I only care that the final uploaded file for me the person reading the logs is available.  Dedent that from `noisy`?  However you want to do it.  It's not a very clear parameter name.
</blockquote></details>

Okay, `upload` is now always noisy.
___
*[`tools/upload-to-aws.py` line 85 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4idNhi8id6CrgZeAaK:-N4mrIDJ1dlfV25OsOX7:b-oekqa5) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L85)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

> Whether it makes sense or not is debatable.

It doesn't, lets remove `stderr`, nothing else uses it.  We aren't beholden to the arbitrary decisions of old.

>  "Pretty" is not a goal here. Precision is.

Good point, that may help debug it one day.
</blockquote></details>

Done. (Er... warnings/errors use `stderr`, but this isn't in that category.)
___
*[`tools/upload-to-aws.py` line 98 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ievet1CKQ_jk2o6l9:-N4mrTcDBACvOz3Su3cR:b-896fix) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L98)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

> It should be obvious from the output that attempts were made

It won't be immediately obvious why the script stopped.  Something like `print(f"ERROR: attempted {max_attempts} unsuccessful uploads.")`.  Makes it searchable, `ERROR` i think makes it show up in cdash.  The `CalledProcessError` tracebacks are somewhat useless, but I'm more interested in a user facing (person searching the logs) error message explaining what went wrong.  keeping `raise` is fine
</blockquote></details>

Done.
___
*[`tools/upload-to-aws.py` line 119 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4ig0aL0kn4hflNyPsB:-N4mrVfE080Ozb9FMXuF:b-896fix) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L119)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

> That requires that the artifact's containing directory is writable ... any desire to make this more general

There isn't, we know with certainty we can write to the directory where the artifact came from in CI.

> I don't see why it's problematic.

The file has not been closed yet when you ask another process to start reading it, that isn't always successful and if it does fail it will be hard to understand what went wrong from CI logs.  Alternative:

```py
fd, path = tempfile.mkstemp(text=True)
os.write(fd, f'{checksum.hexdigest()}  {name}\n')
os.close(fd)
upload(...)
os.remove(path)
```
</blockquote></details>

Done.
___
*[`tools/upload-to-aws.py` line 164 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iTE_e5z7_R7-Jzo8W:-N4mn0V44vzAedGM-lMS:b-tmarqg) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L164)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

OK, cmake side validates it.
</blockquote></details>

Right, any breakage (including exceptions) will be noticed by CMake :slightly_smiling_face:.
___
*[`tools/upload-to-aws.py` line 166 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-N4iT0cZ1vKZyJwIenjH:-N4mmYwl-t4evGNSsBR-:bqdk8ey) ([raw file](https://github.com/robotlocomotion/drake-ci/blob/8a24ec15de48f929425c7d0b4ca5905b89099071/tools/upload-to-aws.py#L166)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

LGTM, `type=str.lower` is very cool :sunglasses: never seen that!
</blockquote></details>

Yeah, that was new to me also. I really wanted `supported_tracks` to be lower-case because that's how the script always uses them, so went looking for "argparse case insensitive choices" and found it :grin:.

svenevs (approved):
:lgtm_strong: it's really nice to have this in python :heart:

https://drake-jenkins.csail.mit.edu/job/linux-focal-unprovisioned-gcc-bazel-experimental-snopt-mosek-packaging/21/

Currently running, lets merge after.  I just looked, the only weekly packaging build is for macOS, but it does **not** upload artifacts.  I'm buildcop right now so I'll be keeping an eye on things, I don't know how to test continuous or nightly in advance without just merging it.

> Reviewed 1 of 5 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-:-N4mu9-p3YKNit_sx8B0:b-z79vil)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

mwoehlke-kitware (commented):
Missed dropping some `noisy`; trying again as build 22... (AFAIK you *can* manually trigger a nightly/continuous build, though I didn't look if they support a CI SHA. Of course, those builds would be "lies"...)

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-:-N4myHqPExMEorysGADm:b-xsiwbp)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @svenevs)

svenevs (approved):
> Missed dropping some noisy; trying again as build 22...

Gah! we both missed it :/ Good thing we're being patient and thorough with the builds.

>  you can manually trigger a nightly/continuous build

Ah you're right, we can replay them and edit the parameters, there's just no "Build with Parameters" thing there.

https://drake-jenkins.csail.mit.edu/view/Continuous%20Production/job/linux-focal-unprovisioned-gcc-bazel-continuous-snopt-mosek-packaging/340

https://drake-jenkins.csail.mit.edu/view/Nightly%20Production/job/linux-focal-unprovisioned-gcc-bazel-nightly-snopt-mosek-packaging/66/

There's nothing mac specific here and I don't think we need to wait for a mac job on this one, I take responsibility and will as build cop swiftly fix anything that comes up next week :slightly_smiling_face:

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/164#-:-N4mzEKH1KdYxE1cIhqh:b-tzsalx)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

svenevs:
Everything checked out on experimental, continuous, and nightly (downloaded and verified checksums too).  Note to future us: replaying a production nightly job is inappropriate, that overwrites `drake-latest-focal.tar.gz`.  This may result in a failure tomorrow if nightlies run tonight, depends on of s3 lets you overwrite an upload or if it has to be deleted.  Either way, changing the artifacts midday shouldn't be done... lesson learned.

PR #167, opened 2022-06-23
Fix jammy package resolution issues with systemd versions

svenevs:
The following packages have unmet dependencies:
   systemd-timesyncd : Depends: systemd (= 249.11-0ubuntu3) but 249.11-0ubuntu3.1 is to be installed
E: Unable to correct problems, you have held broken packages.

We aren't sure if this is actually necessary long term, but opening a PR enables building against it with a separate drake PR.

### DIFF
--- a/setup/ubuntu/install_prereqs
+++ b/setup/ubuntu/install_prereqs
@@ -91,7 +91,7 @@ EOF
 systemctl --now --quiet enable /lib/systemd/system/xvfb.service
 
 apt-get install --no-install-recommends -o Dpkg::Use-Pty=0 -qy \
-  systemd-timesyncd
+  systemd systemd-timesyncd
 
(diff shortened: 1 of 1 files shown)


mwoehlke-kitware:
As I tried to explain on Chat, this won't help; an incompatible `systemd` is *already* installed. Locally, I can no longer reproduce, and the current version of [both](https://packages.ubuntu.com/jammy-updates/amd64/systemd) [packages](https://packages.ubuntu.com/jammy-updates/amd64/systemd-timesyncd) is 249.11-0ubuntu3.3.

I believe Jammy CI is just borked until the repo mirrors are back in sync.

mwoehlke-kitware:
Specifically, it looks like *Amazon* has borked their mirror, and it's been that way for almost 24 hours.

PR #168, opened 2022-06-24
Work around broken EC2 Jammy mirror

mwoehlke-kitware:
Since approximately 2022-06-22, Amazon's EC2 mirror of the Jammy package repository has contained an updated `systemd` (`249.11-0ubuntu3.1`) with an out-of-date and incompatible `systemd-timesyncd` (`249.11-0ubuntu3`). This prevents the latter from being installed, which in turn prevents CI from completing its pre-build setup.

Work around this by explicitly permitting downgrades when installing `systemd-timesyncd`.

Note that, as of 2022-06-24, the current version of these packages is `249.11-0ubuntu3.3`.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/168)

### DIFF
--- a/setup/ubuntu/install_prereqs
+++ b/setup/ubuntu/install_prereqs
@@ -90,8 +90,22 @@ EOF
 
 systemctl --now --quiet enable /lib/systemd/system/xvfb.service
 
-apt-get install --no-install-recommends -o Dpkg::Use-Pty=0 -qy \
+timesyncd_packages=(
+  libsystemd0
+  systemd
   systemd-timesyncd
+)
+
+systemd_compat_version="$(
+    $WORKSPACE/ci/tools/apt-common-version.py "${timesyncd_packages[@]}"
+)"
+
+for ((i = 0; i < ${#timesyncd_packages[@]}; ++i)); do
+    timesyncd_packages[$i]="${timesyncd_packages[$i]}=$systemd_compat_version"
+done
+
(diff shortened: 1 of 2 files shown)


jwnimmer-tri:
Would another possible work-around be to skip the timesyncd install on Jammy entirely?  I'm not sure if it's critical to our operations or not.

mwoehlke-kitware:
> Would another possible work-around be to skip the timesyncd install on Jammy entirely? I'm not sure if it's critical to our operations or not.

@jwnimmer-tri, I suspect it's there to ensure timestamps are reasonable and that the system clock is correct. To some extent, that isn't critical, but I'm not sure how much I want to trust that the VMs clocks are accurate without it. (In particular, I'd be somewhat concerned about e.g. the nightly builds getting versioned "correctly".)

Also, the last attempt indicates that the *cache* thinks it has `249.11-0ubuntu3.3` available, which suggests something is *really* screwed up in the EC2 mirror. Another possible fix, though surely not an ideal one, is probably to just not use that mirror. Otherwise, the only thing I can think to do is force all of the packages to their non-`updates` version (i.e. `249.11-0ubuntu3`). This PR (in its current state) was an attempt at something that could detect the problem and work around it in a way that won't be egregiously "wrong" when and if it ever gets fixed, whereas I wouldn't feel good with permanently hard-coding the non-`update` versions being anything other than temporary.

Right now I'm inclined to hope Amazon gets their act together over the weekend...

PR #170, opened 2022-07-06
Use chrony for NTP, explicitly sync

mwoehlke-kitware:
Switch from `timesyncd` to `chrony` for NTP clock synchronization on Ubuntu. This is for two reasons; first, because `timesyncd` being version-locked to `systemd` has resulted in a number of headaches when EC2's mirrors get out of sync (and have an updated `systemd` but no `timesyncd`), and second because various sources recommend `chrony` over `timesyncd` for virtual machines anyway.

Also add a step before we do anything else to explicitly trigger a synchronization. This was more or less being done on unprovisioned jobs, and should be happening anyway if the daemon is doing its job, but this makes doubly sure and also provides some output that should help verify that the clock is actually correct.


mwoehlke-kitware:
Test builds:
- https://drake-jenkins.csail.mit.edu/job/linux-focal-unprovisioned-gcc-bazel-experimental-release/186/
- https://drake-jenkins.csail.mit.edu/job/linux-jammy-unprovisioned-gcc-bazel-experimental-release/19/

BetsyMcPhail:
Let's make sure we didn't break focal provisioned builds:

BetsyMcPhail (approved):
:lgtm: with a follow-up task to update provisioned images and remove todos

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/170#-:-N6K59F1894bsz32XVfH:b-342fwk)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

PR #180, opened 2022-08-31
[ignore] we are soul searching

svenevs:
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/180)


svenevs:
Fortunately users cannot obtain secret values from us via the jenkins build log, it will get stripped out :+1:

PR #227, opened 2023-06-16
Push Docker staging images

mwoehlke-kitware:
Simplify a bunch of logic. Modify staging jobs to also push Docker images.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/227)


BetsyMcPhail (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/227#-:-NY4eRhgA7dTTxZJKPTy:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

jwnimmer-tri:
The new logic isn't working:

https://drake-jenkins.csail.mit.edu/view/Staging/job/linux-focal-unprovisioned-gcc-bazel-staging-packaging/5/

I'll circle back later to see whether this is a blocker for today's release, or if its innocuous for now.

BetsyMcPhail:
I will look into what is different with this job vs. the Nightly jobs that are allowed to push to docker hub.

BetsyMcPhail:
>  I wonder if $ENV{DOCKER_PASSWORD_FILE} isn't set for the staging job(s)?

That's it, the Staging Jenkins job needs to be updated, working on a fix now.

jwnimmer-tri:
> ... would it be acceptable to re-run the jobs?

It's always fine to run the Staging jobs, you just can't use `1.18.0` as the version number since we already have finished artifacts in place for that build (besides docker).

BetsyMcPhail:
The Jenkins staging jobs have been updated.

mwoehlke-kitware:
Grr, `step-build-docker-image` also needs changes. 😩

mwoehlke-kitware:
> you just can't use 1.18.0 as the version number since we already have finished artifacts in place for that build (besides docker).

...so we won't be able to test those bits until 1.19? Or unless we roll a 1.18.1 for the sole purpose of testing the release process (which IIRC you didn't want to do)?

jwnimmer-tri:
Not sure which "those bits" you're referring to.

You can test the `Staging` job code by using a dummy version number when running the job.

You can test https://github.com/RobotLocomotion/drake/pull/19631 by manually mimicking what the Staging job would have done (by pushing the candidate image manually) and then trying that new code to complete the process.

However, looking more carefully at the output of the Staging jobs I ran (e.g., https://drake-jenkins.csail.mit.edu/view/Staging/job/linux-focal-unprovisioned-gcc-bazel-staging-packaging/5/console), it looks like the tag is wrong anyway?  I thought it was supposed to be a version number, not a date.  I'll dig in more after my next hour of meetings.

mwoehlke-kitware:
> You can test RobotLocomotion/drake#19631 by manually mimicking what the Staging job would have done (by pushing the candidate image manually)

That's not the same as an end-to-end test, though. Also, don't the candidate images disappear when the staging job goes away? (I suppose we could re-tag the corresponding nightly images, which is at least not-worse than we've been doing.)

> looking more carefully at the output of the Staging jobs [...] it looks like the tag is wrong anyway? Ah, maybe that's what #228 is for.

Right, those are the needed `step-build-docker-image` to which I previously alluded.

jwnimmer-tri:
> That's not the same as an end-to-end test, though.

When 1.19.0 comes around next month, we'll find out whether everything works together as a whole.

In the meantime, testing the incremental steps manually will at least detect any bugs in those incremental steps that would surely break the end-to-end test.

> Also, don't the candidate images disappear when the staging job goes away?

No?  I thought the staging outputs were on an S3 cleanup timeout like all other artifacts.

mwoehlke-kitware:
> I thought the staging outputs were on an S3 cleanup timeout like all other artifacts.

I was assuming the instances have no permanent storage, i.e. once the job finishes, anything it didn't explicit push elsewhere is gone. @BetsyMcPhail indicated the same belief. In any case, *I* don't know offhand how to recover the ephemeral state of an already-finished job, and I think @svenevs has already turned into a pumpkin for the day (about that time for me, also). If *you* know how to recover that state and are able to push `{jammy-,focal-,}1.18+staging`, that would allow us to get 'proper' images pushed for 1.18 for the first time. Otherwise that will have to wait for 1.19.

jwnimmer-tri:
Ah, I see, you were asking to recover the bit-exact docker images that the Staging build was using.  You're right that we don't have those anymore.

However, you could recreate something functionally equivalent by taking the staging tgz and re-running the Dockerfile, like I did on slack for the nightlies, and then push that to Dockerhub with the `1.18+staging` tag.

mwoehlke-kitware:
> However, you could recreate something functionally equivalent [...]

Ah. Yes, that might work. My knowledge of how those images get generated is inadequate to have produced that suggestion; thanks.

PR #229, opened 2023-06-20
Fixup to errors in Docker push

BetsyMcPhail:
Add a missing argument to correctly format the docker push command.

Also, properly update the status so that CI will fail on errors.

Follow-up to #227

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/229)

### DIFF
--- a/driver/configurations/bazel/step-push-docker-image.cmake
+++ b/driver/configurations/bazel/step-push-docker-image.cmake
@@ -32,7 +32,7 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-function(docker_push)
+macro(docker_push TAG)
   if(NOT DASHBOARD_UNSTABLE)
     set(DOCKER_PUSH_IMAGE_ARGS "push robotlocomotion/drake:${TAG}")
     separate_arguments(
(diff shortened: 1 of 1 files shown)


mwoehlke-kitware (commented):
:lgtm:, thanks! :flushed:

I normally prefer to use `set(PARENT_SCOPE)` and avoid macros, but it's not obvious how to do that here, and probably not worth it.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/229#-:-NYP_M1L04BiFD5NSYp-:bglad0v)* status: 0 of 1 files reviewed, all discussions resolved

jwnimmer-tri:
> Also, properly update the status so that CI will fail on errors.

It wasn't immediately obvious to me how this patch addresses that problem.  Is it the change from function to macro that somehow affects how errors are processed?

mwoehlke-kitware:
> It wasn't immediately obvious to me how this patch addresses that problem. Is it the change from function to macro that somehow affects how errors are processed?

I assume that's the case. Specifically, it looks like `DASHBOARD_UNSTABLE` wasn't escaping the `function`.

BetsyMcPhail (commented):
Like Matthew mentioned, when the `append_step_status` macro was called from the `docker_push` function, it set the `DASHBOARD_UNSTABLE` result variable in that function's scope only. We need to make `docker_push` a macro (or alternatively rewrite or copy `append_step_status` to use `set(DASHBOARD_UNSTABLE ... PARENT_SCOPE)`) in order to set the global variable outside of the function. Eventually, `DASHBOARD_UNSTABLE` is one of the variables checked to determine if a job succeeded.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/229#-:-NYPafwP0h8Ts5Efjf-O:b83g7rv)* status: 0 of 1 files reviewed, all discussions resolved

jwnimmer-tri:
Got it, thanks.

mwoehlke-kitware:
`append_step_status` shouldn't assume it's called from a function. Perhaps it could set the variable in *both* local and parent scopes, but that seems kludgy. It would be more "correct" for `docker_push` to push `DASHBOARD_UNSTABLE` to parent scope, but that's also a bit awkward. Hence my original review, that it probably isn't worth trying to keep `docker_push` a `function`. 🙂

BetsyMcPhail (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/229)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @BetsyMcPhail)

PR #242, opened 2023-09-03
WIP rewrite

jwnimmer-tri:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242)

### DIFF
--- a/tools/pip_index_url.py
+++ b/tools/pip_index_url.py
@@ -1,181 +1,137 @@
 """Create an ``index.html`` suitable for use with ``pip --index-url``.
 
-Scrape which nightly wheels are available (not in glacier storage) from drake's nightly
-s3 bucket.  Create PEP 503 https://peps.python.org/pep-0503/ compliant ``index.html``
-and uploaded it to the drake s3 bucket.
+Scrape which nightly wheels are available (not in glacier storage) from drake's
+nightly s3 bucket.  Create PEP 503 https://peps.python.org/pep-0503/ compliant
+``index.html`` and uploaded it to the drake s3 bucket.
 
-This script runs automatically every morning at 9:00am eastern so that the nightly
-artifacts are complete.  The script can also be run manually to regenerate as needed
-(e.g., nightly builds were failing and it needs to re-run after they were fixed).
+This script runs automatically every morning at 9:00am eastern so that the
+nightly artifacts are complete.  The script can also be run manually to
+regenerate as needed (e.g., nightly builds were failing and it needs to re-run
+after they were fixed).
 
-NOTE: to develop locally, you must have your ~/.aws/config and ~/.aws/credentials
-already configured for drake.
+NOTE: to develop locally, you must have your ~/.aws/config and
+~/.aws/credentials already configured for drake.
 """
 from __future__ import annotations
 
+import datetime
+import io
+import operator
 import re
 import sys
-import tempfile
+from dataclasses import dataclass
 from pathlib import Path
 from textwrap import dedent
 
 import boto3
 
 
+@dataclass
+class Wheel:
+    s3_key: str
+    yyyymmdd: str
+    py_minor: int  # e.g. the 11 in "Python 3.11"
+    sha512: str = None
+
+
 def main() -> None:
-    # Accumulate any errors and display at the end.
-    error_messages: list[str] = []
-
-    # Keys: wheel basename, Values: sha512 hash
-    drake_nightly_wheels: dict[str, str] = {}
-
-    # Login to s3 and query the drake-packages bucket for wheel files not already
-    # transitioned to glacier storage.  Do not include any "drake-latest" wheels such
-    # as `drake-latest-cp310-cp310-manylinux_2_31_x86_64.whl`, we only want to include
-    # actual versioned wheel files.
-    #
-    # Since the sha512 files are uploaded separately, we will download each of those
-    # in order to include the expected sha512 hash of the wheel file.
-    temp_directory = tempfile.TemporaryDirectory(prefix="drake_pip_index_url_")
-    temp_directory_path = Path(temp_directory.name)
-
-    # It is helpful to include logging to the console so the user understand the
-    # script is still working on things.
-    print("==> Logging into s3.")
+    # Log in to s3.
+    print("==> Logging in to s3 ...")
     s3 = boto3.resource("s3")
     bucket_name = "drake-packages"
-    drake_nightly_bucket = s3.Bucket("drake-packages")
-    print(f"==> Querying {bucket_name} objects.")
-    for obj in drake_nightly_bucket.objects.all():
-        if (
-            obj.storage_class == "STANDARD"
-            and obj.key.startswith("drake/nightly")
-            and obj.key.endswith(".whl")
-        ):
-            # Overwrite drake/nightly/<some path>.whl with just <some path>.whl
-            obj_basename = Path(obj.key).name
-            if not obj_basename.startswith("drake-latest"):
-                print(f"- FOUND: {obj.key}")
-                sha512_name = f"{obj.key}.sha512"
-                print(f"  DOWNLOAD: {sha512_name}")
-                try:
-                    sha512_obj = s3.Object(
-                        bucket_name=bucket_name,
-                        key=sha512_name,
-                    )
-                    # NOTE: use the basename of `sha512_name` to avoid needing to
-                    # create drake/nightly/ subdirectories.
-                    sha512_dest = temp_directory_path / Path(sha512_name).name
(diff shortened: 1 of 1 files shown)


svenevs (changes requested):
Overall I'm a big fan of these changes.  I'm going to fiddle around a little more with how the actual layout is supposed to work (solve the `NotImplementedError`).  But I'm also OK to merge this as-is into the other branch and keep working there.  The comments here don't really need to block.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-:-NdcMyz0EgiPg_N4u-A5:b-boh3h0)* status: all files reviewed, 5 unresolved discussions (waiting on @jwnimmer-tri and @svenevs)

___
*[`tools/pip_index_url.py` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-NdcMCaZALDuqGPhqDHU:-NdcMCa_-x_6CPmqwDcj:br9ez9l) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/5d9b76b20e4ec13f633bf767d38642b45b1f421f/tools/pip_index_url.py#L49)):*
> ```Python
>     wheels: list[Wheel] = []
>     print(f"==> Querying {bucket_name} objects ...")
>     days_back = 48
> ```

working, we can probably go longer?  These are the lifecycle rules for `drake/nightly` 



But, we may want to go shorter.  One of the things worth looking into is whether or not `pip` will cache this index.  If they do, we may not be able to solve anything
___
*[`tools/pip_index_url.py` line 74 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-NdcM420D1s4xBICFZsp:-NdcM420D1s4xBICFZsq:b-u176q0) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/5d9b76b20e4ec13f633bf767d38642b45b1f421f/tools/pip_index_url.py#L74)):*
> ```Python
> 
>     # Sort by newest.
>     wheels.sort(key=operator.attrgetter("yyyymmdd", "py_minor"), reverse=True)
> ```

nit, there's no _need_ to sort, right?  I like it, it makes it coherent.  Just making sure the logic below doesn't actually depend on it.
___
*[`tools/pip_index_url.py` line 83 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-NdcL3CpDhY8I-1ltVfG:-NdcL3CpDhY8I-1ltVfH:bmc9vfo) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/5d9b76b20e4ec13f633bf767d38642b45b1f421f/tools/pip_index_url.py#L83)):*
> ```Python
>         sha_data = io.BytesIO()
>         s3.Object(bucket_name, sha_key).download_fileobj(sha_data)
>         wheel.sha512 = sha_data.getvalue().decode("utf-8").strip().split()[0]
> ```

BTW NICE!
___
*[`tools/pip_index_url.py` line 109 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-NdcK07fD50x_WW4pRk_:-NdcK07fD50x_WW4pRka:bhp4l9p) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/5d9b76b20e4ec13f633bf767d38642b45b1f421f/tools/pip_index_url.py#L109)):*
> ```Python
>         html.write(
>             f'<a href="https://{server}/{s3_key}#sha512={sha512}" '
>             f'data-requires-python="&gt;=3.{py_minor},&lt;3.{py_minor+1}">'
> ```

working, I'm not sure if this is the right thing to do.  I don't see any reason why `&lt;3.{py_minor+1}` would be an issue, and probably this is a `drake` packaging issue, but we aren't uploading the metadata that would imply this to PyPI.  View source for the webpage https://pypi.org/simple/drake/ there are only `>3.8` really.
___
*[`tools/pip_index_url.py` line 114 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-NdcLGDt9EFjlMNx5HJA:-NdcLGDt9EFjlMNx5HJB:b699rs1) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/5d9b76b20e4ec13f633bf767d38642b45b1f421f/tools/pip_index_url.py#L114)):*
> ```Python
> 
>     print(html.getvalue())
>     raise NotImplementedError()
> ```

working, to be implemented (@svenevs)

jwnimmer-tri:
Yeah, I figured you could merge this to your branch (with some amendments both during and/or after) and we can keep going from there.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242)* status: all files reviewed, 5 unresolved discussions (waiting on @svenevs)

___
*[`tools/pip_index_url.py` line 49 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-NdcMCaZALDuqGPhqDHU:-NdcNeJ1A1ZHb0FSAURq:b-vkbmz9) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/5d9b76b20e4ec13f633bf767d38642b45b1f421f/tools/pip_index_url.py#L49)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

working, we can probably go longer?  These are the lifecycle rules for `drake/nightly` 



But, we may want to go shorter.  One of the things worth looking into is whether or not `pip` will cache this index.  If they do, we may not be able to solve anything
</blockquote></details>

I figured the last few days didn't really matter.  This was giving up to ~1 week for any local caching as a ballpark.
___
*[`tools/pip_index_url.py` line 74 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-NdcM420D1s4xBICFZsp:-NdcNljg6LKBohcG7czs:b-xdall2) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/5d9b76b20e4ec13f633bf767d38642b45b1f421f/tools/pip_index_url.py#L74)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

nit, there's no _need_ to sort, right?  I like it, it makes it coherent.  Just making sure the logic below doesn't actually depend on it.
</blockquote></details>

Right, it's mostly just so that the scripts actions are always deterministic.

Also it puts the newest wheels first, for anyone browsing the HTML by hand in their browser that seemed nicest.   Though now I see that PyPI puts oldest on top and newest at the bottom, so maybe we should not use the the `reverse=True` part.
___
*[`tools/pip_index_url.py` line 109 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-NdcK07fD50x_WW4pRk_:-NdcO59EA6CbtfFXraJ9:bu1vta0) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/5d9b76b20e4ec13f633bf767d38642b45b1f421f/tools/pip_index_url.py#L109)):*
<details><summary><i>Previously, svenevs (Stephen McDowell) wrote…</i></summary><blockquote>

working, I'm not sure if this is the right thing to do.  I don't see any reason why `&lt;3.{py_minor+1}` would be an issue, and probably this is a `drake` packaging issue, but we aren't uploading the metadata that would imply this to PyPI.  View source for the webpage https://pypi.org/simple/drake/ there are only `>3.8` really.
</blockquote></details>

Huh, interesting.  I didn't dig too deeply to see exactly which tools assume what.  But the extra annotation is _true_ right?  Maybe just pointless?

svenevs (approved):
:lgtm: ok sounds good, i'll merge it and annotate the other PR with the things here.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-:-NdhFTXJ44ZtV72Ktuf1:b-xuwoob)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jwnimmer-tri)

___
*[`tools/pip_index_url.py` line 109 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/242#-NdcK07fD50x_WW4pRk_:-NdhFc1o8wHXoVHhNkKZ:b-3h7ki5) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/5d9b76b20e4ec13f633bf767d38642b45b1f421f/tools/pip_index_url.py#L109)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

Huh, interesting.  I didn't dig too deeply to see exactly which tools assume what.  But the extra annotation is _true_ right?  Maybe just pointless?
</blockquote></details>

done, AFAICT it's the right thing to do, if anything the drake pypi uploads need to be fixed.  I'll stress test it when we get there

PR #251, opened 2023-11-09
Use release version in staging tarballs

mwoehlke-kitware:
Modify how we build tarballs to use the release version rather than the date for staging builds.


mwoehlke-kitware:
Since `download_release_candidate.py` is sort-of pulling the nightly binaries ATM (n.b. https://github.com/RobotLocomotion/drake/issues/20339#issuecomment-1804711931; in particular, we *are* using nightlies for macOS), I think it's "safe" to go ahead and merge this even if updating `download_release_candidate.py` is WIP.

PR #275, opened 2024-03-28
Fix/improve error reporting

mwoehlke-kitware:
Always use append_step_status to communicate errors rather than setting the variables in situ. Don't treat Debian archive problems as both UNSTABLE and FAILURE, as this prevents the error from being communicated via the final status report. (Nothing downstream runs if the build is UNSTABLE, so treating these as FAILUREs is unnecessary.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/275)


BetsyMcPhail (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/275)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

PR #289, opened 2024-09-03
Add apt hold for shim-signed on Jammy

BetsyMcPhail:


BetsyMcPhail (commented):
+@mwoehlke-kitware for review. I'm undecided if this is the best solution. I updated the unprovisioned Jammy AMI with the change below and then added it to this script so we don't lose it if the AMI gets recreated from scratch. The other option would be to leave the AMI alone and update https://github.com/RobotLocomotion/drake-ci/blob/main/setup/ubuntu/install_prereqs, where some similar holds already exist.... Do you have any preference?

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/289#-:-O5sTV1q3TmkAW--9F1m:b-k4wzrs)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

mwoehlke-kitware (commented):
:lgtm:, after looking into what this package is, it looks like it should be pretty harmless to keep it at the same version as the unprovisioned image. Can't say I'm aware of any particular logic for what should go in `ami_init_script` vs. `install_prereqs`, so... :shrug:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/289#-:-O61aFbuALcKiUoEc6TV:bngxwjq)* status: 0 of 1 files reviewed, all discussions resolved

BetsyMcPhail (commented):
+@tyler-yankee for review please

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/289#-:-OPaa1Er1ayOtsYz25y8:bcxtrv7)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @mwoehlke-kitware and @tyler-yankee)

tyler-yankee (commented):
Note that when I created new images this past week for https://github.com/RobotLocomotion/drake/issues/22890, I didn't actually start from scratch for Linux unprovisioned; I took the existing unprovisioned and uninstalled old Java / installed new Java. So, I don't have any actual verification of whether or not this workaround is still required.

If this is needed for all images, then the code should be in this script, otherwise, it can go with the rest of the holds in `install_prereqs`. I tend to think we're in the former?

I guess I'm not understanding why this workaround was needed, why it's only temporary, and whether or not we can get rid of it now given that we didn't _technically_ make new unprovisioned images from scratch.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/289#-:-OPad8FtBVa5vN7H8WYW:bojfca0)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @mwoehlke-kitware)

BetsyMcPhail (commented):
This is similar to #330 in that it is a fix to a CI issue. In this case, however, the fix was applied directly to the image and not via a setup script. You're right, it's not needed unless the Jammy unprovisioned image is created completely from scratch. On the other hand, it should have lived in a script to begin with and not just applied to the image.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ci/289#-:-OPap0Vl-frt0jE-fIok:b-ac3hc6)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @mwoehlke-kitware and @tyler-yankee)

PR #306, opened 2025-03-04
Remove hard-coded server from pip index

jwnimmer-tri:
The index file is necessarily on the same server.

This is relevant for when our main DNS entry goes down, to allow our other CNAME records for the same data to still serve requests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/306)


jwnimmer-tri:
+@mwoehlke-kitware do you have cycles to review this, please?

PR #313, opened 2025-03-27
Adjust macOS packaging logic

tyler-yankee:
Towards [#22819](https://github.com/RobotLocomotion/drake/issues/22819). With the transition to AWS, macOS packaging jobs will now be provisioned. This means we need special logic for them not to use the remote cache, as before.

In addition, macOS packages are only currently being published/uploaded on Sonoma, not Sequoia. This logic can be updated now.


tyler-yankee:
<details><summary><i>Previously, BetsyMcPhail (Betsy McPhail) wrote…</i></summary><blockquote>

Would  a test on the existing experimental Mac packaging job be worthwhile to make sure this didn't accidentally break something? I can't think of a way to test on Mac AWS.
</blockquote></details>

I'll kick off an experimental Mac packaging job. Mac on AWS essentially already has this logic, since it never uses the remote cache and I made the change to publish Sequoia independently on that branch.

tyler-yankee:
<details><summary><i>Previously, tyler-yankee (Tyler Yankee) wrote…</i></summary><blockquote>

I'll kick off an experimental Mac packaging job. Mac on AWS essentially already has this logic, since it never uses the remote cache and I made the change to publish Sequoia independently on that branch.
</blockquote></details>

[`mac-arm-sonoma-unprovisioned-clang-cmake-experimental-packaging`](https://drake-jenkins.csail.mit.edu/view/Packaging/job/mac-arm-sonoma-unprovisioned-clang-cmake-experimental-packaging/27/) succeeded, uploaded results to S3, and didn't use the remote cache. But it's still unprovisioned, and still Sonoma, so I'm not sure how much value that test has.

BetsyMcPhail (changes requested):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/313)* status: all files reviewed, 2 unresolved discussions (waiting on @tyler-yankee)

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ci/313#-OMRFcgH0CbkkA-BdM0f:-OMS4m1p1jUCGpffNh7w:bvpd9r2) (no related file):*
<details><summary><i>Previously, tyler-yankee (Tyler Yankee) wrote…</i></summary><blockquote>

Maybe spin up a quick mac-arm-sequoia-clang-cmake-experimental-packaging (non-AWS) and try to run it with this branch?
</blockquote></details>

Thanks, I really just wanted CI to confirm that we didn't miss a silly typo somewhere.
___
*[`driver/environment.cmake` line 141 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/313#-OMS790j93APyKFb0-kT:-OMS790j93APyKFb0-kU:bc7d3yw) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/e9141ac1e11596d05577f0762d52009e621ff682/driver/environment.cmake#L141)):*
> ```CMake
> set(REMOTE_CACHE ON)
> 
> if(DASHBOARD_JOB_NAME MATCHES "(health-check|unprovisioned)" OR
> ```

On second thought, could `packaging` just be added to the original regex here? We don't want on caching on any packaging jobs and that would make this comparison simpler and easier to understand.

tyler-yankee:
*[`driver/environment.cmake` line 141 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/313#-OMS790j93APyKFb0-kT:-OMSAINF1qeyEsrwcqWh:bjt0u0) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/e9141ac1e11596d05577f0762d52009e621ff682/driver/environment.cmake#L141)):*
<details><summary><i>Previously, BetsyMcPhail (Betsy McPhail) wrote…</i></summary><blockquote>

On second thought, could `packaging` just be added to the original regex here? We don't want on caching on any packaging jobs and that would make this comparison simpler and easier to understand.
</blockquote></details>

Oh, I didn't realize that; then yes, much simpler.

tyler-yankee (commented):
@BetsyMcPhail Do you know of any other updates that need to happen on the Drake side (e.g. perhaps `release_engineering`) if we're going to start publishing Sequoia packages? That may be worth looking into.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/313#-:-OMSHU5a3wQfQFWk0X39:b-oubzrj)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @BetsyMcPhail)

jwnimmer-tri:
> ... to start publishing Sequoia packages?

Only "experimental" jobs upload to S3, right?  In that case, no further documentation is necessary.

Ah, I see the code here looks wrong.  Only Sonoma Nightly should be "publish"; the Sequoia should still only be just "ON".

BetsyMcPhail (commented):
The Mac AWS jobs currently use an experimental drake-ci branch (https://github.com/RobotLocomotion/drake-ci/commits/aws-provisioned-test/). This branch contains the logic to not publish the Sequoia nightly packages, and some other Mac AWS - specific code.  At this point, maybe we're ready to de-experimentalize the branch and merge those changes into drake-ci? 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/313#-:-OMTZJ6g-bg5zaniIcLy:b-785ffd)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @tyler-yankee)

jwnimmer-tri:
Ah.  Yes, having two different branches in scope for everyday use is confusing.  If we need different logic for the AWS test, we should have like "if AWS" in the main drake-ci cmakelists, normal review process, etc.

BetsyMcPhail (commented):
Makes sense, Tyler will have a better idea of the work required to merge the AWS branch.

I will note that Jenkins only has Sonoma packaging jobs at the moment. Regardless, it sounds like we should explicitly enforce the Sonoma-only limitation in the drake-ci scripts. 

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/313#-:-OMTdJz8CJawLZ_6BIBQ:bvulw2)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @tyler-yankee)

jwnimmer-tri:
> I will note that Jenkins only has Sonoma packaging jobs at the moment. Regardless, it sounds like we should explicitly enforce the Sonoma-only limitation in the drake-ci scripts.

I see.  If the new thesis is "we will only ever define nightly packaging jobs using correct os version", then I'm OK removing the extra checks here.  I hadn't connected the dots that DJJ was where our leverage was coming from these days.

jwnimmer-tri:
Maybe a good middle ground here is to have the `drake-ci` just assert that any macOS packaging job is running on Sonoma.  That still lets us delete the complicated if-else logic, but still provides a second line of defense in case we make a mistake in DJJ definitions.

tyler-yankee:
A couple thoughts:

* Merging the aws-provisioned-test branch into main isn't that difficult, but it will require a corresponding DJJ change to update the `-aws` jobs to use main instead.
* > ...assert that any macOS packaging job is running on Sonoma.
 
  I'm OK with doing it this way, or just leaving it as-is / reverting the changes on this PR. Either way, it's going to require immediately modifying the logic for the AWS case.

tyler-yankee:
Okay, let's do the following (after offline discussion with @BetsyMcPhail):

* Leave the existing packaging logic for Sonoma and Sequoia alone. meaning Sequoia packages can be built, but not published
* That means the only change for this PR is to have the packaging jobs not use the remote cache.
* We can merge this PR first, and then the PR on DJJ to remove the unprovisioned jobs and change the packaging jobs to provisioned.
* Finally, I can make follow-up PRs here and on DJJ to merge the AWS test branch into main and have all the existing jobs use it. I suppose this also goes with creating nightly and staging packaging jobs on AWS.

jwnimmer-tri:
SGTM

BetsyMcPhail (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/313#-:-OMhV7pfBMeJ21vLxfKJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tyler-yankee)

PR #317, opened 2025-04-08
Add more verbose resource usage logs

tyler-yankee:
Currently, we log disk usage before and after builds. It would also be useful to know memory usage -- how much is being used before/after, and by which processes? We can use this to identify extraneous processes running on CI machines that can be turned off to optimize build performance, and/or determine optimal build/test concurrency with Bazel.

(And while we're at it, since we use `ps` to get the top processes for memory usage, include CPU usage information for those processes as well. If the it's pertinent information, I'm willing to add a separate command which actually *sorts* by CPU usage so we can see top consumers there too, but figured what's here is a good start for now.)

CI examples of what these logs would look like:
* Linux Jammy: https://drake-jenkins.csail.mit.edu/job/linux-jammy-gcc-cmake-experimental-release/6001/
* macOS Sonoma: https://drake-jenkins.csail.mit.edu/job/mac-arm-sonoma-clang-cmake-experimental-release/22/

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/317)

### DIFF
--- a/ctest_driver_script.cmake
+++ b/ctest_driver_script.cmake
@@ -121,8 +121,8 @@ set(DASHBOARD_CDASH_URL "")
 set(ENV{CMAKE_CONFIG_TYPE} "${DASHBOARD_CONFIGURATION_TYPE}")
 set(CTEST_CONFIGURATION_TYPE "${DASHBOARD_CONFIGURATION_TYPE}")
 
-# Report disk usage before build
-execute_step(common report-disk-usage)
+# Report resource usage before build
+execute_step(common report-resource-usage)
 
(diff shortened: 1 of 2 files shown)


BetsyMcPhail (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/317#-:-ONv1rnR9E51mD4bTgRI:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tyler-yankee)

PR #319, opened 2025-04-16
Add explicit Bazel build server idle timeout

tyler-yankee:
We've had occasional macOS build failures in AWS with OOM errors as a result of running back-to-back builds on the same machine. Bazel's default [build server idle timeout](https://bazel.build/reference/command-line-reference#startup_options-flag--max_idle_secs) is 3 hours, so the server is sticking around from previous builds and causing high memory pressure. Since we only need the server to stay alive in between CMake calls to Bazel, keeping only a couple minutes should be sufficient.


tyler-yankee:
+@BetsyMcPhail for review, please.

tyler-yankee:
Note CI results with `--max_idle_secs=99` included for both CMake and Bazel builds:
* https://drake-jenkins.csail.mit.edu/job/mac-arm-sonoma-clang-cmake-experimental-release/23/
* https://drake-jenkins.csail.mit.edu/job/mac-arm-sonoma-clang-bazel-experimental-release/2854/

PR #332, opened 2025-05-13
[cache] Rotate log files more frequently

tyler-yankee:
The nginx `access.log` files are building up in large excess of the 100 MB set in `logrotate_cache.conf` (up to triple the size in some cases, which is significant when we have 10 rotations). The solution here is to rotate them more often --- every 15 minutes, instead of hourly --- so that the size gets caught before blowing up too much. We also offset it from the other two cron jobs so that the log files don't clash.

Also, amends c1061a349ab1860ca2fa86a2f2780db9c8093989 to update the threshold in the cron job for monitoring `/` with the same value as is present in `health_check.bash`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/332)

### DIFF
--- a/cache_server/README.md
+++ b/cache_server/README.md
@@ -218,12 +218,12 @@ All of the configuration options should be executed as `root`.
     #
     # Additionally monitor disk usage of the root volume.  This is where the
     # logging data is stored.
-    40 * * * *   /opt/cache_server/drake-ci/cache_server/disk_usage.py / >>/opt/cache_server/log/disk_usage_root.log 2>&1
+    40 * * * *   /opt/cache_server/drake-ci/cache_server/disk_usage.py / -t 70 >>/opt/cache_server/log/disk_usage_root.log 2>&1
     #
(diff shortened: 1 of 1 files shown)


tyler-yankee:
+@BetsyMcPhail thoughts here? can you think of anything else -- perhaps being more strict on the size of the log files in `logrotate_cache.conf`? As of writing, the `access.log` files go back a few days, which I think is probably a good amount if we ever needed to debug and wanted history.

BetsyMcPhail (approved):
I'm good with giving this a try and seeing how it goes. 

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/332#-:-OQAKNmo7nyisI6B1qdi:b-6z2jm6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tyler-yankee)

PR #316, opened 2025-04-04
Transition macOS CI logic to AWS

tyler-yankee:
Towards [#22858](https://github.com/RobotLocomotion/drake/issues/22858), this completes the transition from MacStadium to AWS ... almost[^1].

* Removes any logic associated with `-aws` in the job name, and either replaces with `mac` if it needs to stick around or deletes the functionality because it was only temporary.
* Removes `mac-arm64` logic (and associated documentation) from `health_check.bash`, as we'll no longer be using the cache server at MacStadium.

[^1]: "Almost" because of the above mentioned cache server transition, which will still have to be done in a future PR if we intend to host a cache server for macOS in AWS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/316)

PR #335, opened 2025-05-19
Allow macOS jobs to use the remote cache

tyler-yankee:
Closes https://github.com/RobotLocomotion/drake/issues/22997. We disabled the remote cache for all macOS jobs in #316 as part of the transition to AWS, since the Mac Stadium cache server would no longer be in use. Here, we'll try pointing it to the same cache server as Linux uses.

If we end up using this long-term, it might be worth cleaning up some of the code and comments in `cache.cmake` and `health_check.bash`, since the general design there was to have arguments/options for multiple cache servers.

### DIFF
--- a/cache_server/logrotate_cache.conf
+++ b/cache_server/logrotate_cache.conf
@@ -1,4 +1,4 @@
-# Rotate logs if they grow past 100MB, or each month, whichever is first.
+# Rotate logs if they grow past 150MB, or each month, whichever is first.
 # Note that the /opt/cache_server/log/nginx/access.log in particular grows very
 # quickly.  Logs are rotated hourly as a result, otherwise you can end up with
 # an access.log of >5GB.
@@ -10,6 +10,6 @@
     monthly
     missingok
     rotate 10
-    size 100M
-    maxsize 100M
+    size 150M
+    maxsize 150M
 }
--- a/driver/configurations/cache.cmake
+++ b/driver/configurations/cache.cmake
@@ -318,8 +318,6 @@ if(REMOTE_CACHE AND Java_JAVA_EXECUTABLE)
 endif()
 
 
-# NOTE: As of 4/10/25, the remote cache is disabled on all macOS jobs,
-# but this logic will be left here for if/when it is set up again for AWS.
 if(APPLE)
   if(REMOTE_CACHE)
     set(DASHBOARD_OS_CACHE_NAME "macos")
--- a/driver/environment.cmake
+++ b/driver/environment.cmake
@@ -131,9 +131,7 @@ endif()
 
 set(REMOTE_CACHE ON)
 
-# NOTE: As of 4/10/25, the remote cache is disabled on all macOS jobs
-# as they're hosted in AWS.
-if(DASHBOARD_JOB_NAME MATCHES "(health-check|unprovisioned|packaging)" OR APPLE)
+if(DASHBOARD_JOB_NAME MATCHES "(health-check|unprovisioned|packaging)")
   set(REMOTE_CACHE OFF)
 endif()
 


tyler-yankee:
CI job: https://drake-jenkins.csail.mit.edu/view/Mac%20Sequoia/job/mac-arm-sequoia-clang-bazel-experimental-release/25/

I verified by watching the remote cache access logs that it's issuing a bunch of GET requests, but they're all giving 404s. I assume we would need to let this bake for a few days so that macOS jobs actually upload to the cache, and then jobs could pull from it?

tyler-yankee:
+@BetsyMcPhail for review, please.

tyler-yankee:
*[`cache_server/logrotate_cache.conf` line 13 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/335#-OQdiSYM8k8PQXoeQSeI:-OQdiSYM8k8PQXoeQSeJ:b-scz6o5) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/d0cf04e4c5f84138538bb848e144c9a08a3a0335/cache_server/logrotate_cache.conf#L13)):*
> ```
>     missingok
>     rotate 10
>     size 150M
> ```

This value is purely just a guess. Currently our log files go back ~3 days and disk usage on `/` is 70%. So I figured this change allows us to preserve _some_ history while not overloading the disk.

Definitely open to suggestions here (maybe we move logs to `/cache` in a separate move? Or a follow-up if needed?)

tyler-yankee:
*[`driver/configurations/cache.cmake` line 324 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/335#-OQdj7yW0meIwt35Ur8R:-OQdj7yW0meIwt35Ur8S:b-luajn5) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/d0cf04e4c5f84138538bb848e144c9a08a3a0335/driver/configurations/cache.cmake#L324)):*
> ```CMake
>   if(REMOTE_CACHE)
>     set(DASHBOARD_OS_CACHE_NAME "macos")
>     find_program(SW_VERS_EXECUTABLE NAMES "sw_vers"
> ```

I'm very surprised that the above CI job actually worked given this code is wrong ...

tyler-yankee:
*[`driver/configurations/cache.cmake` line 324 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/335#-OQdj7yW0meIwt35Ur8R:-OQdjmeAC7rHd2Bt-I-A:b7jva19) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/d0cf04e4c5f84138538bb848e144c9a08a3a0335/driver/configurations/cache.cmake#L324)):*
<details><summary><i>Previously, tyler-yankee (Tyler Yankee) wrote…</i></summary><blockquote>

I'm very surprised that the above CI job actually worked given this code is wrong ...
</blockquote></details>

Actually, is it? It ends up leading to how the key is computed below ... wouldn't we _want_ macOS jobs to have their own keys so they get separated from the Linux files?

tyler-yankee:
*[`driver/configurations/cache.cmake` line 324 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ci/335#-OQdj7yW0meIwt35Ur8R:-OQsz6VACXs6OfwAx-mb:b-3vzo30) ([raw file](https://github.com/RobotLocomotion/drake-ci/blob/d0cf04e4c5f84138538bb848e144c9a08a3a0335/driver/configurations/cache.cmake#L324)):*
<details><summary><i>Previously, tyler-yankee (Tyler Yankee) wrote…</i></summary><blockquote>

Actually, is it? It ends up leading to how the key is computed below ... wouldn't we _want_ macOS jobs to have their own keys so they get separated from the Linux files?
</blockquote></details>

From f2f: yes, currently this would mean Ubuntu and Mac jobs compute a different key so they keep their cache files separate. This isn't _strictly_ necessary as Bazel does some of this under the hood, but we have it set up due to different versions of packages/etc. on provisioned images, and the hashes are computed more conservatively than they need to be. Let's just leave it for now.

tyler-yankee (commented):
(ping @BetsyMcPhail)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/335#-:-ORRGZpe340vNwahCcc9:b-fzs8q3)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @BetsyMcPhail)

PR #349, opened 2025-09-04
[cache] Move log files to `/cache/log/`

tyler-yankee:
742074a was still not good enough. `/cache` has significantly more storage, so use it instead of the small `/`.

This introduces a small risk that if the cache file removal fails causing the disk to fill up, we lose the logs. However, that is unlikely, and the logs are likely unhelpful in that case anyways (especially with often they were being rotated before).

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ci/349)


tyler-yankee:
Deployed. I fetched this PR branch and followed the instructions of the README to update the crontab entries and restart nginx. Once merged, I'll check out 'main' again.

